PR #1, opened 2018-04-05
GPU AML - mali update to r7p0 - Subtitles fix - improve AML rendering

wrxtasy:
- Update Mali drivers to r7p0 thx to @kszaq 
- Fix S912 external .srt subtitles video stuttering by disabling Kodi dirty regions completely
- And as a consequence I'm now seeing much improved CPU software decoding performance and rendering on a AML S912 U9. 
- Should be good enough for smooth 1080p H264 Netflix and Amazon Video streaming.

Untested with AML Kodi Leia.

### DIFF
--- a/packages/graphics/opengl-meson/package.mk
+++ b/packages/graphics/opengl-meson/package.mk
@@ -30,9 +30,11 @@ case $MESON_FAMILY in
     ;;
   gxbb)
     if [ "$TARGET_ARCH" = "arm" ]; then
-      PKG_VERSION="8-r5p1-01rel0-armhf"
+      PKG_VERSION="8-r7p0-01rel0-armhf"
+      PKG_SHA256="d79526d849c2cc0c513335b7f78b9ea568c04d241af3a8c4636298a3ae4c9211"
     else
-      PKG_VERSION="gxbb-r5p1-01rel0"
+      PKG_VERSION="gxbb-r7p0-01rel0"
+      PKG_SHA256="9391b7a6863b5d34105e6ba2cc06a489eed0ca3f0b80aa3c2aa3dddebda7b747"
     fi
 ;;
 esac
--- a/packages/linux-drivers/gpu-aml/package.mk
+++ b/packages/linux-drivers/gpu-aml/package.mk
@@ -32,7 +32,7 @@ PKG_LONGDESC="gpu-aml: Linux drivers for Mali GPUs found in Amlogic Meson SoCs"
 PKG_TOOLCHAIN="manual"
 PKG_IS_KERNEL_PKG="yes"
 
-PKG_UTGARD_VERSION="r5p1"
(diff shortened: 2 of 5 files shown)


adamg88:
Thanks for the PR @wrxtasy 

This was put into the latest devel build to get some feedback from forum users.

boot2k3:
This pull doesn't have any regression comments on Leia test builds for S912. I think it can be added.
But I don't know if algorithmdirtyregions (kodi: advancedsettings.xml disable Dirty Regions completely to fix Subtitles) change required by default, it can be added as additional setting.

afl1:
opengl-meson-t82x: 3b21f97  has some performance degradation on  4k HDR 60fps for text overlay . I tested it on high bitrate samples by Samsung and LG (https://drive.google.com/uc?export=download&id=0Bxj6TUyM3NwjNFNmTUJEVjZjMDA with Ctrl-Shift-O)

wrxtasy:
>opengl-meson-t82x: 3b21f97 has some performance degradation on 4k HDR 60fps for text overlay . I tested it on high bitrate samples by Samsung and LG

I see the exact same Kodi fps performance fps degredation no matter which Chipset, set of GPU drivers or advancedsettings.xml used.

For example even 1080p without r7p0 will have video frame upsets when I pull up the OTA TV overlay.
It's a Kodi issue IMHO.

Also S912 CoreELEC is missing a needed libhybris patch I believe:
https://github.com/kszaq/LibreELEC.tv/blob/libreelec-8.2/packages/devel/libhybris/patches/libhybris-fbdev_window.patch

adamg88:
@wrxtasy the libhybris patch is not needed for Leia as the patch was superseded and the changes upstreamed by kszaq before he left the scene.

I can't speak for the other issues as however osmc also tried the r7p0 changes and reverted back after regressions it caused.

wrxtasy:
> the libhybris patch is not needed for Leia as the patch was superseded and the changes upstreamed by kszaq before he left the scene.

Please check the libhybris package hashes again, nothing from that patch was upstreamed by kszaq

We are going to have to agree to disagree it seems and that's fine.

adamg88:
@wrxtasy it wasn't libhybris that the changes where upstreamed too it was Kodi hence I said the patch was superceeded, I will try to find the commit and link back

adamg88:
@wrxtasy https://github.com/xbmc/xbmc/pull/13377

wrxtasy:
Ahh thx. upstream xbmc. was looking  in the wrong place.

DavideFioravanti:
@wrxtasy I compiled from your fork and later from master adding your patches but I didn't see any performance increase. Same poor performance in Netlifx even in 720p. How did you get the performance increase?
Thanks.

Raybuntu:
There is no point leaving this open unnecessarily long. We don't see any benefits.

PR #27, opened 2018-04-22
Add missing buttons to mecool remote

host505:
Following the procedure described [here](https://forum.libreelec.tv/thread/11643-le9-0-remote-configs-ir-keytable-amlogic-devices/?pageNo=1) for a meccol remote control ([this](https://img.sunsky-online.com/upload/store/detail_l/EAT0420_7.jpg)) I found some buttons missing from the built-in config:
0x22 is the bottom left button, left from '0' and 
0x18 is the 'browser' button, the one with an 'e' on it.
Not sure about the commands I assigned, couldn't think of something better.


host505:
Ok I saw you added key 0x18 on this [commit](https://github.com/CoreELEC/CoreELEC/commit/037da6758c71a05c808a904eb3e62bbffcb5be17), but why not 0x22 as well?

PR #41, opened 2018-04-27
rockchip patches

alorbach:
Fix building failure with rock64 and fixed h264 playback crashes with newer rkmpp version.


alorbach:
Yes, you can add my personal email al@x-sites.de to the list. I can contribute to Rockchip development for CoreElec as far as I can. This type of projects is not my area of expertise to be honest ;).

PR #43, opened 2018-04-28
projects/Amlogic: update DTB upgrade priority

adamg88:
I'm sending this change as a PR as it is a significant change to the update process and I would like to get others opinions on it.

There is a lot of users who ignore instructions given for how to update and manually place DTB images in the update folder before updating and this prevents them from receiving the latest DTB.

If no DTB is placed in the update folder then the update process will flash /usr/share/bootloader/$LE_DT_ID.img over the current DTB however if a DTB is placed in the update folder then this does not occur.

I have changed this so *.dtb is ignored in the update folder and the update process can only be overridden if the DTB is named dtb.override.img in the update folder so we can still test DTB changes for development purposes easily.

### DIFF
--- a/projects/Amlogic/bootloader/update.sh
+++ b/projects/Amlogic/bootloader/update.sh
@@ -23,6 +23,7 @@
 [ -z "$UPDATE_DIR" ] && UPDATE_DIR="/storage/.update"
 UPDATE_DTB_IMG="$UPDATE_DIR/dtb.img"
 UPDATE_DTB=`ls -1 "$UPDATE_DIR"/*.dtb 2>/dev/null | head -n 1`
+UPDATE_DTB_OVERRIDE_IMG="$UPDATE_DIR/dtb.override.img"
 [ -z "$BOOT_PART" ] && BOOT_PART=$(df "$BOOT_ROOT" | tail -1 | awk {' print $1 '})
 if [ -z "$BOOT_DISK" ]; then
   case $BOOT_PART in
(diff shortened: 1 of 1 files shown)


arthur-liberman:
I like the idea. One question though.
What happens if the user is on a really old dtb that doesn't have le-dt-id? Then his dtb will not be updated at all. Can/Should the script be changed so that if le-dt-id doesn't exist in the current dtb, the dtb.img is accepted?

adamg88:
This is something I really wanted to prevent as it is the root cause of a lot of issues, even now many weeks after the change from amremote to meson-ir there is still the odd user that tries to follow the instructions to setup a remote and will get the 'no device found' message and you/I know straight away why this is and that they are solely responsible for this as they have ignored advice and copied an old DTB they have saved to their desktop during the update process.

There was conflicting advice on the forums telling people that the dtb and zip must be placed in the update folder during updating when it hasn't been necessary to do this for some time and no matter how many times I repeat myself it is ignored.

To continue allowing the same behaviour is counter-intuitive and there would be no point in me opening a PR for it then.

This PR consists of 3 changes.

Ignore .update/dtb.img
Ignore .update/*.dtb
Use if exists .update/dtb.override.img (purely to ease development purposes)

adamg88:
Actually @arthur-liberman I have just re-read your response and see what you mean, you mean to change the current priority so /usr/share/bootloader/$LE_DT_ID.img is preferred?

arthur-liberman:
I understand what you're saying.
What I'm suggesting is:
Use if exists .update/dtb.override.img (purely to ease development purposes)
Use le-dt-id entry, ignore dtb.img even if it exists.
Only use dtb.img if le-dt-id was not found in current dtb.

adamg88:
That's actually a good idea, don't know why I didn't think of that myself =D

Ok PR updated, behaviour now is as follows...

1. .update/dtb.override.img will be first choice
2. /usr/share/bootloader/$LE_DT_ID.img will be second choice
3. .update/dtb.img will be third choice
4. .update/*.dtb will be fourth/final choice

PR #58, opened 2018-05-02
fix hauppauge in amlogic-3.14

gelotus:
In new media_tree 2018-04-15 package there is a new mt9t112 driver (i2c cam) that requires devm_gpiod_get_optional() function. There's no devm_gpiod_get_optional() in 3.14. Simply remove it


gelotus:
Uh.. yeah that's much better

gelotus:
Fixed by @afl1

PR #66, opened 2018-05-05
Update version

jlazkano:


gelotus:
For now we try to stay in sync with LE master. In future we'll see. Btw thanks for your interest in CE

PR #86, opened 2018-05-11
Rename FD628 to OpenVFD.

arthur-liberman:
This is just the first step. 
I think that separating the add-on into its own repository is a good idea, and will make development easier.
The 'fd628-aml' folder will have to be renamed, and possibly moved out of amlogic if we're going to build it for Rockchip too.


arthur-liberman:
Cancelling this PR and opening a PR to merge into master.

PR #91, opened 2018-05-15
config/packages: remove os.libreelec dependency

adamg88:
There is a bug within Kodi that seems to not recognize the os.libreelec.tv dependency within addon's even though it exists on the system.

This gives the false impression to the users that they are missing a dependency when they are not.


juanesf:
I have a dependency problem in coreELEC 8.90.2, but it's called os.libreelec.tv

adamg88:
just press OK and the addon will install as normal

PR #100, opened 2018-05-26
Python2: update to Python-2.7.15

### LINKED ISSUE #99: Project does not build with host gcc8 due to python2.7.14 defect
Src: https://bugs.python.org/issue33374

Moving to python2.7.15 fixed it for me.
Is there any reason to stay on python2.7.14?
If not i could prepare a pull request that updates python2 and 3 to their latest versions.

MartB:
This fixes #99 
Refreshed patches and added now working `--enable-optimizations` option. (https://bugs.python.org/issue28032#msg275182)

### DIFF
--- a/packages/lang/Python2/package.mk
+++ b/packages/lang/Python2/package.mk
@@ -17,8 +17,8 @@
 ################################################################################
 
 PKG_NAME="Python2"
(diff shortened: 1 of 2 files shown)


adamg88:
Thanks

PR #104, opened 2018-05-29
NSS/NSPR update to 3.37.1/4.19

MartB:
This is required to get coreelec to build with host gcc8.
(Also updating this was long overdue)

Tested and running on ODROID-C2

nss-06-always-include-fstar.patch is an upstream defect, it should be fixed with the next release.


MartB:
tested working with widevine and the latest inputstream.adaptive changes.

PR #119, opened 2018-06-09
Change default kodi resolution to 1080p

boot2k3:
Sometimes after reboot or power on kodi didn't receive resolutions from TV and set it by default to 720p.
After this change default resolution is 1080p


adamg88:
Has this been tested?

boot2k3:
It's not so simple to repeat, 3rd day without issues.
Leave it open, may be somebody have this issue too and can check.
I will report after week. I had this issue 2-3 times a week before.

PR #123, opened 2018-07-06
Add support for installing to internal with LVM

adamg88:
This PR is self explanatory, the idea is to maximise available disk space when installing to internal.

/flash reduced to 512mb, the same as all other LE supported devices
/storage increased by 1.5gb

The increase in storage is achieved by reducing the size of the /dev/system partition and combining /dev/cache and /dev/instaboot with LVM.

This does not interfere with USB booting or users who are installed to internal with the usual method.


adamg88:
Abandoning as installing to internal is discouraged.

PR #125, opened 2018-07-12
kodi: disable SSL certificate validation

adamg88:
A couple of users are having issues with using https with self-signed SSL certificates, this can be worked around by adding verifypeer=false to the URL but this PR negates the need to do this.

Bad practice obviously.

I think another work around would be to move SSL certificates to /storage/.config/ssl from /etc/ssl and symlink it back and request users add their certificate to the PEM file.


arthur-liberman:
IMO it is a bad idea to hamper security with hacks like this. 
If there's a workaround that will allow users to use their self signed certificates and still keep SSL secured for everybody else, we should go that route.

PR #129, opened 2018-08-10
pull request

computerbastler:


computerbastler:
pull request

computerbastler:
update

Raybuntu:
There are 2 Problems with this. First of all we don't know about the legal implications. Netflix is a registered Trademark and also it is probably against their ToS.
2nd we would have to maintain it and keep it up to date. I'd rather include the NF authors repo. But since it's available from Kodinerds repo which we already preinstall I don't see a point.

Raybuntu:
Rule Nr. 1: One PR per feature
Rule Nr. 2: Drop the unnecessary Merge Commits
Rule Nr. 3: Proper commit names and comments
Rule Nr. 4: Don't remove SHA256 when bumping Packages.

PR #133, opened 2018-09-07
RTC shield support improvement

PabloCamp0s:
These commits provide an out of the box for the Odroid C2 RTC shield (of course, rtc_shield='1' is still needed into /flash/config.ini).

First commit addresses any dtb status, and sets the "okay" or "disabled" properly. I've found that my dtb starts from "disabled", which is not handled correctly atm.

The second time moves pcf8563 into kernel. As a module, I'm getting this into dmesg:

```
[    0.461945@0] drivers/rtc/hctosys.c: unable to open rtc device (rtc0)
[    4.559254@1] rtc-pcf8563 1-0051: chip found, driver version 0.4.3
[    4.562308@1] rtc-pcf8563 1-0051: rtc core: registered rtc-pcf8563 as rtc0

```
If pcf8563 is compiled into kernel I'm getting this:

```
[    0.310181@1] rtc-pcf8563 1-0051: chip found, driver version 0.4.3
[    0.310906@1] rtc-pcf8563 1-0051: rtc core: registered rtc-pcf8563 as rtc0
[    0.464073@1] rtc-pcf8563 1-0051: setting system clock to 2018-09-07 15:13:39 UTC (1536333219)

```

(note the "setting system clock" part).

"hwclock" command executions are already provided by the existing udev rule /lib/udev/rules.d/80-clock.rules so I think this PR is complete right now.

BTW, thanks all the maintainers for all your hard work.


Raybuntu:
What happens is no rtc shield is present. We manage rtc shield support through config.ini. Means disable is by default in boot.ini and enable through config.ini. Otherwise the log gets spammed with errors

PabloCamp0s:
rtc_shield is still equals to 0 in boot.ini, therefore the default behaviour is disabled. User must set rtc_shield='1' into config.ini to enable the shield.

Raybuntu:
That's intended, so what are you saying?

PabloCamp0s:
@Raybuntu, you said: 

> What happens is no rtc shield is present.

If this is a question, I haven't tested with no rtc shield plugged. However, if this is the case, the default is set to disabled the "/i2c@c1108500/pcf8563@51".

> We manage rtc shield support through config.ini. Means disable is by default in boot.ini and enable through config.ini.

I know. This behaviour has not been changed whatsoever.

> Otherwise the log gets spammed with errors

In what case do we get error spamming?

Thanks.

Raybuntu:
I apologise if I sound rude. This change was made because 99% of C2 users have no rtc shield and we see a lot of kernel log errors which affect system performance.

Raybuntu:
Before we hat rtc shield enabled in dtb and it cause regression through kernel log spam.
I don't think changing to built-in solves it. As I said you don't have issues when the rtc shield is present vut that's 1% of all Users.

The "okay" dtb commit is actually what's bothering me.

PabloCamp0s:
> The "okay" dtb commit is actually what's bothering me.

But putting the "okay" only happens if the user puts rtc_shield='1' into the config.ini. Otherwise boot.ini works exactly as it's doing right now. The code in boot.ini is

`if test "${rtc_shield}" = "1"; then fdt set "/i2c@c1108500/pcf8563@51" status okay; else fdt set "/i2c@c1108500/pcf8563@51" status disabled; fi`

If rtc_shield is 0 (default), the code always puts the "disabled".

Raybuntu:
Ahh I know what you mean. This is actually a bug cause RTC shield was enabled by default in LE source
https://github.com/LibreELEC/linux-amlogic/blob/amlogic-3.14.y/arch/arm64/boot/dts/meson64_odroidc2.dts#L834
We don't have that. That's why I was confused.
You can drop the "else" then cause it's disabled by default.

PabloCamp0s:
> You can drop the "else" then cause it's disabled by default.

Agreed. But I think that the else part is more failsafe and manages better the desired user configuration (I want the RTC, let's put the okay... I don't want it, ok then, let's put the disabled).
Besides, [this PR made by @adamg88](https://github.com/CoreELEC/CoreELEC/pull/43) several months ago created a new logic for loading custom DTBs and, therefore, putting the default value of the property to an unknown one. With the else part, even such cases are covered.

Raybuntu:
Good

Raybuntu:
Is it necessary to make the module built-in?

Raybuntu:
Ok. Can you also try removing the rtc and see if there are any log messages spamming with this change when rtc_shield is disabled.

PabloCamp0s:
Done.

dmesg with rtc_shield = '1' in config.ini and no shield plugged, you're right, tons of i2c messages. However, this is an expected outcome.
```
[   28.797586@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.797612@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.797680@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.797695@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.797757@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.797773@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.797799@2] rtc-pcf8563 1-0051: pcf8563_set_datetime: err=-11 addr=02, data=21
[   28.797874@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.797890@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.797959@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.797974@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.798037@2] i2c i2c-1: [aml_i2c_xfer] error ret = -5 (-EIO)
[   28.798052@2] i2c i2c-1: token 1, master_no(1) 300K addr 0x51
[   28.798065@2] rtc-pcf8563 1-0051: pcf8563_get_datetime: read error
```
dmesg with rtc_shield variable unset in config.ini and shield unplugged has no i2c messages at all.

Looks good to me.

PR #138, opened 2018-10-26
kodi, pvr.iptvarchive: Add support for IPTV Archive features

arthur-liberman:
This PR adds a modified pvr.iptvsimple add-on and the accompanying modifications to Kodi to support it.
It adds support for the Archive feature some IPTV services provide to their users.
The add-on adds the following improvements over the standard pvr.iptvsimple add-on:
- Support for playing a show from the EPG, using the archive.
- Practically unlimited timeshift support.
- Lets the user select a show from the EPG and timeshift to it.

### DIFF
--- a/packages/mediacenter/kodi-binary-addons/pvr.iptvarchive/package.mk
+++ b/packages/mediacenter/kodi-binary-addons/pvr.iptvarchive/package.mk
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (C) 2018-present Team CoreELEC (https://coreelec.org)
+
+PKG_NAME="pvr.iptvarchive"
+PKG_VERSION="c7c136afca9414808b35b4d73a365300b8af9436"
+PKG_SHA256="0b244677d80ef69d4f25e741e8c1195bdfe4734dc1d15d3419daa441e3b1f3e3"
+PKG_REV="2"
+PKG_ARCH="any"
+PKG_LICENSE="GPL"
+PKG_SITE="http://www.kodi.tv"
+PKG_URL="https://github.com/kodi-pvr/pvr.iptvsimple/archive/$PKG_VERSION.tar.gz"
+PKG_DEPENDS_TARGET="toolchain kodi-platform rapidxml zlib"
+PKG_SECTION=""
+PKG_SHORTDESC="pvr.iptvarchive"
+PKG_LONGDESC="pvr.iptvarchive"
+
+PKG_IS_ADDON="yes"
+PKG_ADDON_TYPE="xbmc.pvrclient"
+
+PKG_CMAKE_OPTS_TARGET="-DRAPIDXML_INCLUDE_DIRS=$(get_build_dir rapidxml)"
+
+post_patch() {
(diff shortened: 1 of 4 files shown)


vpeter4:
Fine by me.

I really hope one day we see this functionality in Kodi/pvr.iptvsimple itself.

MartijnKaijser:
Sure being able to record and store **any** IPTV streams available doesn't ring any alarm bells. Not to start over patching Kodi itself as well.

arthur-liberman:
@MartijnKaijser 
First of all this change does NOT record, nor store any streams. All it lets you do is start playback from the specified timestamp. Using the same mechanism, it also allows the user to "seek" anywhere they please in the timeline.
I am aware that a custom Demux component was the better way to go, but Kodi has pretty much everything needed to get it working, but the support for it is just not there yet.
This feature, at this point is specific to CoreELEC, and I don't have plans to upstream it as is.

Raybuntu:
@MartijnKaijser Team Kodi proved several times that it'd like to play Source code Police with comments like "Raping my code" etc. This is still open source and we will not be intimidated to NOT patch Kodi. You guys can relicense closed source that will solve your problem. There is no reason to play bullie just because YOU can.

As Arthur explained there is no reason so go and bother another Kodi fork like OSMC .

vpeter4:
This small kodi patch brings BIG and much, much, much better Kodi user experience for some users. Including me. And I like it very, very much! Timeshifting is something "must include". Of course even play something from EPG in pvr.iptvsimple is something. But better have 2 functionality :)
Wish Kodi would include something like that but then I wake up in the middle of the nigh soaked in sweat :)

MartijnKaijser:
If this feature makes sense to have in the true upstream code then it should be PRed to that. If CoreELEC decides to go diverge too much than yes they will have to completely rebrand. Just like OSMC already did so that's the reason we will not bother them. @Raybuntu So you argument is invalid. 

@arthur-liberman Just to be clear we are open to any features that make sense and fit our design, and some one is willing to do the work for it.

Raybuntu:
@MartijnKaijser We already have a bunch of non-upstreamed and non-upstreamable Patches/Hacks so does LibreELEC, Debian and other forks. So in a sense this is already diverged a lot.
But what is your problem anyway? Comments like this and other comments just make upstreaming pointless.
OSMC is not a real rebrand according to your trademark policy. Binary is still called kodi.bin for instance and log is called kodi.log.
If that's acceptable it's an easy fix for us when we fork a skin that doesn't say kodi or has any logo's.

I don't see a point in rebranding because 
1.) We don't use any Kodi resources (forum, bugtracker, personal Team Kodi support) like others do.
2.) We don't add piracy stuff like others do and we don't break Kodi. Even if when we use hacks we use them to improve Kodi so there is no reason for Team Kodi to claim that we harm the brand.
3.) We only have approx. 2000 users, we don't make any money like OSMC does with sales or LibreELEC with Donations. So using the name Kodi does no benefit to us and falls into Fair use in our case.

But I will no longer discuss this in this PR here, if Team Kodi has a Problem or wants to shut us down they can write us an email. I'm just tired of being constantly harassed by Team Kodi Members.

PR #145, opened 2018-12-03
projects/Amlogic: add support for NanoPi K2

adamg88:
This PR adds support for NanoPi K2.

Unfinished.. u-boot still needs patching to accept cfgload.


Raybuntu:
I'm against this. For one simple reason. Neither of us have that box and I don't wanna support yet another u-boot. 
At the least it should use our AML u-boot @Portisch is working on so we don't have yet another u-boot.

PR #146, opened 2018-12-07
Add a multithreaded build system

arthur-liberman:
This PR allows us to build as many package as possible in parallel.
It brings huge build time savings on machines with high processor counts.


Raybuntu:
Awesome buildtime went down from 1h20min clean S905 build to 22.5min!
All Addons build in 15min now including slow mono.

PR #147, opened 2018-12-11
Fix amlogic libhybris compile problem under arch aarch64

phorcys:
libhybris configure.ac use arm64 to detect arch
CoreELEC use aarch64


Raybuntu:
We only build "arm" packages so we didn't notice this. I'm ok changing this. Thanks

PR #148, opened 2018-12-17
Adding UVC camera support for Amlogic

goolpone:
Adding UVC camera support enabling kernel driver.


goolpone:
It would be useful to have UVC camera support on CoreELEC

Raybuntu:
Thank you. I have nothing against this if it works and doesn't break something else.

adamg88:
Closing as no response, feel free to open the PR again if you wish to update it.

goolpone:
> This PR is missing the config entry for SOC_CAMERA

SOC_CAMERA entry is intended for cameras not connecting over a bus like PCI or USB (For example some i2c camera connected directly to the data bus of an SoC), but on Amlogic box, no cameras are provided if not the one externally attached to a USB. So UVC driver support for cameras would be enough.

goolpone:
@adamg88 You're right! I did not notice this entry. Fixed on #153

PR #150, opened 2018-12-20
fix Wake-on-Lan by external RTL8211F for S912 devices

Portisch:
This patch will use the PMEB interrupt pin of the RTL8211F device to check
for a matching WOL packet. The PMEB pin have to be connected to GPIOZ_15.


Portisch:
Wrong repo

PR #152, opened 2018-12-31
lepotatoemmc: toggle eMMC timeout earlier

kszaq:
eMMC timeout should be toggled earlier when installing as we don't want to have the timeout enabled in config.ini copy on eMMC.


Raybuntu:
Thanks @kszaq. Nice to see you again :)

PR #153, opened 2019-01-11
Adding UVC camera support for Amlogic

goolpone:
Adding UVC camera support enabling kernel driver.


adamg88:
Can you rebase these into a single commit and force push the change to update the PR

git rebase -i HEAD~2

Change to fix up for second commit and then

git push yourremote -f

goolpone:
Ok, rebased as requested, thanks for the hint!

adamg88:
Thanks for the change and the PR

PR #154, opened 2019-01-16
amremote: add service to toggle between meson-ir/remote

adamg88:
as discussed previously meson-ir gives a very poor user experience with bundled remotes with generic devices

this service will toggle between meson-ir and amremote for users who wish to use Krypton based remote configurations


boot2k3:
It's working correct on my S912 box

adamg88:
Thanks for testing @boot2k3

PR #159, opened 2019-01-28
projects/Amlogic: disable hotplug governor

kszaq:
Hotplug governor should not be used for 2 reasons:
- it causes crashes
- it moves irqs back to CPU0 when CPUs1-3 go to sleep.


adamg88:
thanks

PR #160, opened 2019-01-30
Tweak shaders for AML ARM platforms

wrxtasy:
Kodi PR#15286 improved rendering performance:
https://github.com/xbmc/xbmc/pull/15286

Further tweaks are needed for AML ARM platforms.
As suggest by @kszaq in the above PR.

Noticeable improvements, especially for S912 LE/CE users include:

1) 480/576i mpeg2 SW decoding and YADIF2x deinterlacing is now possible.
2) 720p h264 SW decoding and smooth playback when using Kodi Leia's Netflix Addon is now possible
3) No video playback / external subtitle stuttering observed now with S912 LE / CE

### DIFF
--- a/packages/mediacenter/kodi/patches/amlogic/kodi-le-062-xbmc_col_conversion_if_0.patch
+++ b/packages/mediacenter/kodi/patches/amlogic/kodi-le-062-xbmc_col_conversion_if_0.patch
@@ -1,0 +1,25 @@
+From 1375e3db97e9d2256f970831417b1459bab951cc Mon Sep 17 00:00:00 2001
+From: wrxtasy <wrxtasy@amnet.net.au>
+Date: Sun, 20 Jan 2019 07:38:43 +0800
+Subject: [PATCH] xbmc_col_conversion_if_0
+
+---
+ system/shaders/GLES/2.0/gles_yuv2rgb_basic.frag | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/system/shaders/GLES/2.0/gles_yuv2rgb_basic.frag b/system/shaders/GLES/2.0/gles_yuv2rgb_basic.frag
+index ecd82b1..a8345f8 100644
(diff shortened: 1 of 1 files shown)


adamg88:
Thanks @wrxtasy

adamg88:
Thanks for the heads up

adamg88:
@wrxtasy we have had reports that this does not solve the subtitle situation

wrxtasy:
@kszaq did not notice anything, but I have also been testing fritsch's colorspace conversion improvements.

Did a back to back test and i cannot seen any differences to be honest. Will need to throw more SW decoded SD content at it methinks. I suspect it will be pretty rare finding any BT.601 video content to test.

Would be safer leaving such colorspace conversion enable for legacy video support.

Also updated the S912's t82x GPU drivers to your #3b21f97 and I swear during testing the GUI is a bit snappier vs what CoreELEC is currently using.

@adamg88 - only briefly tested. Will have another look. I am testing slightly diff OpenGLES S912 GPU drivers as well. (v3.2)

adamg88:
@wrxtasy [this](https://discourse.coreelec.org/t/coreelec-8-99-1/3341/11) is one users comments who is quite familiar with the subtitle issue, I'm sure there was a reason we didn't update the GPU drivers but I can't remember off the top of my head

wrxtasy:
The included advancedsettings.xml file in the last PR might have caused issues for non S912 users.

With rendering improvements now included in Kodi Leia updating AML S9xx OpenGLES drivers might be worth revisiting. 
Poor Kodi Leia ARM rendering performance I believe has been the main cause of numerous issues I've seen since Kodi Jarvis/Krypton LE releases.

fritsch:
The above patch is wrong and makes no sense at all. Furthermore: Why do you guys still track patches?

adamg88:
@fritsch we are using your powrevert now, I think everybody tracks patches.

fritsch:
Everyone that cannot use git most likely :-)

Am Do., 31. Jan. 2019, 17:00 hat adamg88 <notifications@github.com>
geschrieben:

PR #161, opened 2019-02-01
Update S912 - Android OpenGL ES to v3.2

wrxtasy:
3rd times is a charm...
Definitely fixes S912 external subtitles stuttering. 
Thoroughly tested this time, I did not even clock increase the GPU MHz.

Updated Android Binaries found over on the Freaktab forum.
Source long since forgotten.

For those reading v3.1+ produces a really noticeable increase in rendering performance when using the Android OS 6.x onwards and the T820 GPU due to advanced ARM rendering tech:
https://forum.kodi.tv/showthread.php?tid=330585

This is a dirty patch for testing in the wider S912 CE community.

A better option would be to clone @kszaq 's T820x Repo and patch over the top of commit #915cb48
Then of course change T820 Repo source.


adamg88:
thanks, I will make a build for feedback from some of our users who have issues with subs

adamg88:
that commit only applies to S905, it does nothing for S912

gpu-aml is used on both S905 and S912

boot2k3:
@wrxtasy , could you try this 2 files from archive on your box?
https://dropmefiles.com/DlCu1
Looks much better, but I still see stutter on 1st subtitle

adamg88:
> S912 SW decoded Netflix definitely needs r7p0 for the updated v3.2 Android GLES drivers.
@wrxtasy that is just wrong sorry, you need to look at the make_target more closely in `packages/linux-drivers/gpu-aml/package.mk` because even If we apply that commit S912 will still not use r7p0, only S905 will use it

wrxtasy:
@adamg88 you are correct that I'm wrong. ✌️
Found the cause of my SW decoded 720p Netflix playback. 

Caused by still compiling with this un-needed Linux Kernel patch from:
https://github.com/kszaq/LibreELEC.tv/commit/e16f0f66c43254ce8e694aa7dde3924904b622f7

All good to go with this PR then, unless you want to clone kszaq t820x Repo and patch the Android Open GLES the binaries.

@boot2k3  - the problems with S912 were front what I've seen - external .srt type subtitles. 
These are now fixed with this v3.2 PR.
Tested both your files and sample.mkv is fine, the other file has no subtitles at all.

BTW: the utgard r7p0 kszaq PR's increase SW decoded rendering performance on S905's like an O'clocked  C2.
It's very nearly capable of the lowest 1080p Netflix SW decoding now.
That will be for a seperate PR.

boot2k3:
@wrxtasy , 2nd file also have subtitles, just enable them in subtitle menu.
Could you share you S912 build for test it on my box?

kszaq:
After some thinking patching libhybris to provide GLES3 is pointless for now as Kodi will not benefit from it (yet? https://github.com/xbmc/xbmc/pull/13457).

The vsync patches might be worth a try.

Perhaps S912 might also benefit a little bit from using UNPACK_ROW_LENGTH, you can try using this patch (only for S912!) and SW decode a DVD to check if it works:
https://kszaq.libreelec.tv/temp/s912_kodi_unpack_row_length.patch

adamg88:
Please resend this to coreelec-9.0 branch

PR #162, opened 2019-02-11
Unify S905 and S912 projects

kszaq:
I've been playing with and idea to have a unified build for S905 and S912 for some time now and it looks like thanks to some Kodi linking changes it's finally possible. Kernel has been the same for some time, mali module loads automatically, the only thing that needs solving is GLES library - this can be hacked-around by bind mounting libMali if device is S905.

I don't know much about modifications to the build system CE introduced so I'm only showing my patches that were *not* tested much - fresh installs, upgrades, subdevices all need to be verified.

Feel free to pick and fix or drop. 😃


adamg88:
This is great, we spoke about doing something like this on Slack a couple of months ago when Ray did the initial work on unifying the many different builds into just 2. To take this one step further and unify everything into a single build would be awesome for both users and devs alike.

kszaq:
Branch force-pushed to fix libMali bind mount.

adamg88:
I've tested this on S905 and it works great, still to test S912.

One tidbit I would like to see addressed is the img/tar naming, ie the output was.
```
CoreELEC-Amlogic.arm-9.0-devel-xxxx-Amlogic.img.gz
CoreELEC-Amlogic.arm-9.0-devel-xxxx.tar
```

I was thinking that maybe PROJECT.ARCH should be dropped and Amlogic changing to Generic?

Both can be done with changes in `scripts/image` to IMAGE_NAME and `SUBDEVICES+=" Generic"`.

If you or others can think of another alternative or disagree, I'm all ears.

Raybuntu:
This looks awesome. I would however use a dedicated device rather than the project name so we can add S905x2 soon. Maybe S9xx?

Raybuntu:
On second thought, we should probably consider adding all future SoC's to a common generic AML device. So this is fine with me :+1:

adamg88:
Confirmed working on S912 as well, this is god level awesome!

vpeter4:
What about using some symbolic linking instead of mount/bind? I think it would look better.

kszaq:
This approach is the easiest but I agree that a script creating symlinks in `/var` (like for kernel modules) would look much nicer.

projects/Amlogic/bootloader/canupdate.sh
```diff
@@ -2,7 +2,7 @@
 # Copyright (C) 2018-present Team CoreELEC (https://coreelec.tv)
 
 case $1 in
-   @PROJECT@)
```
Raybuntu: This should not be required

Raybuntu:
I agree with @vpeter4 that we should do it like the kernel module overlays. Also I would check for something in the dtb rather than check if there are more then 4 cpu cores.

projects/Amlogic/bootloader/canupdate.sh
Raybuntu: Sorry I see now

adamg88:
@Raybuntu I don't believe dtb is the best way because both kvim and odroid use different values than standard in compatible and amlogic-dt-id, checking core count works and is probably the easiest.

Raybuntu:
I see.

projects/Amlogic/initramfs/platform_init
```diff
@@ -92,3 +92,9 @@ echo 8 > /proc/irq/62/smp_affinity   # dwc_otg, dwc_otg_hcd:usb2, dwc_otg_pcd
 echo 4 > /sys/module/amvdec_h264/parameters/dec_control
 
 echo 1 > /sys/devices/system/cpu/cpufreq/ondemand/io_is_busy
+
+# S912: Set 1GHz minimum CPU speed for both clusters
+if [ -e /sys/devices/system/cpu/cpu4 ] ; then
```
adamg88:
```suggestion
if [ -f /proc/device-tree/t82x@d00c0000/compatible ]; then
```

might be better for when/if we add newer Amlogic SoC's

projects/Amlogic/packages/opengl-meson/system.d/libmali.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Choose libMali for the device
+Before=kodi.service
+
+[Service]
+Type=oneshot
+ExecStart=-/usr/bin/sh -c '[ ! -e /sys/devices/system/cpu/cpu4 ] && mount -o bind /usr/lib/libMali.so /usr/lib/libEGL.so && mount -o bind /usr/lib/libMali.so /usr/lib/libGLESv1_CM.so && mount -o bind /usr/lib/libMali.so /usr/lib/libGLESv2.so'
```
adamg88:
```suggestion
ExecStart=-/usr/bin/sh -c '[ ! -e /proc/device-tree/t82x@d00c0000/compatible ] && mount -o bind /usr/lib/libMali.so /usr/lib/libEGL.so && mount -o bind /usr/lib/libMali.so /usr/lib/libGLESv1_CM.so && mount -o bind /usr/lib/libMali.so /usr/lib/libGLESv2.so'
```

or maybe look at @vpeter4 / @Raybuntu suggestion? and move to a helper script rather than a oneliner in a service?

projects/Amlogic/packages/opengl-meson/package.mk
```diff
@@ -0,0 +1,23 @@
...
+PKG_TOOLCHAIN="manual"
+
+makeinstall_target() {
+  mkdir -p $INSTALL/usr/lib
+    cp -PR usr/lib/libMali.so $INSTALL/usr/lib
+}
+
+post_install() {
+  enable_service libmali.service
+}
```
adamg88: maybe overwrite the existing package instead in packages/graphics/opengl-meson?

spycat88 (changes requested):

Raybuntu:
@kszaq I've made some minor changes to your PR: See here https://github.com/CoreELEC/CoreELEC/compare/coreelec-9.0...Raybuntu:kszaq-unify_S905_S912?expand=1

This works with symlinks now. Could you rebase your branch? I'm not allowed to push to this PR because it's disabled.

kszaq:
@Raybuntu The branch has been force-pushed to match yours. I have not tested it.

Raybuntu:
Thanks. We tested it team internally. Great work and we really appreciate your support and contribution!

adamg88:
Awesome PR thanks, will save all of us a lot of time

Power535:
PROJECT=Amlogic DEVICE=S905 ARCH=arm make image
ERROR: You need to specify a valid device for the Amlogic project
Please examples for 9.0 and 9.2

boot2k3:
For old devices like S905.. S912:
PROJECT=Amlogic make image

Power535:
Sorry , I do not understand
PROJECT=Amlogic ARCH=arm make image
How separate for commands for ?
1) LePotato
2) Odroid_C2
3) S905
4) S912
Please examples for 9.0 and 9.2

boot2k3:
As I said before just
PROJECT=Amlogic make image
As you can see from this PR description you don't need to compile images separately..
As a result of command you receive Generic image for S905 and S912, LePotato and Odroid_C2 images.

Ray-future:
You need to do a clean build. Otherwise it will cache the old projects.

Power535:
Simply interesting
What this command DEVICE=S905 and support this command ?
Support 9.0 and 9.2 S905 for aarch64 ?

arthur-liberman:
You don't need DEVICE anymore. The build process will create all images.

Power535:
Me need know for diffirents versions including old versions
I am trying to replace kernel
Please message

arthur-liberman:
You need to know what? I don't understand what you are asking.

Power535:
What this command DEVICE=S905 and support this command ?
Support 9.0 and 9.2 S905 for aarch64 ?

arthur-liberman:
CoreELEC uses aarch64 kernel with 32bit userspace.
You don't need to use `DEVICE` at all. I don't know how else to explain it.

Power535:
Possible in future support DMT HDMI 
for kernel 3.14 and 4.9 or exist patch ?
Need 1024 x 768 85 Hz
I have only old minitors
Thank you

PR #163, opened 2019-02-13
libMali-450 update to r7p0, improvement for T820

kszaq:
This is https://github.com/LibreELEC/LibreELEC.tv/pull/2512 adapted for CE.

Improvement on S905 is probably a placebo.

Improvement on S912 is noticeable: SW-decoded DVD stutter with full Yadif, after this patch only a few frames skipped.


projects/Amlogic/patches/linux/linux-001-PR94.patch
```diff
@@ -0,0 +1,92 @@
...
+ 		"display_mode_default", &str);
+@@ -2529,9 +2531,10 @@ static int osd_probe(struct platform_device *pdev)
+ 	else
+ 		current_mode = vmode_name_to_mode(str);
+ 	prop = of_get_property(pdev->dev.of_node, "pxp_mode", NULL);
+-	if (prop)
++	if (prop) {
+ 		prop_idx = of_read_ulong(prop, 1);
+-	osd_set_pxp_mode(prop_idx);
++		osd_set_pxp_mode(prop_idx);
```
arthur-liberman: Shouldn't the changes in this file be fixed in the kernel source rather than just patching it?

Raybuntu: Yes I agree this should go to the kernel

arthur-liberman: Ok, I'll push the fix for it when I get off work.

kszaq: Keep in mind that if you apply the second patch, you need to update mali libs or you will have vsync not working.

arthur-liberman: Ok, but the change above looks like a bug in the kernel that should be fixed, regardless of whether we take this PR or not.

projects/Amlogic/packages/opengl-meson-t82x/package.mk
```diff
@@ -2,8 +2,8 @@
 # Copyright (C) 2017-present Team LibreELEC (https://libreelec.tv)
 
 PKG_NAME="opengl-meson-t82x"
-PKG_VERSION="915cb48"
-PKG_SHA256="9b5f65afa21250b67578c250da030a5829e69131ce91b2f167b01b1ed30be781"
+PKG_VERSION="3b21f97"
+PKG_SHA256="bb97c408a7b802592cb952c8ac117e810635366ab4447c5a335511fc279f3852"
```
shantigilbert: I get checksum ```f2f928b77e87d2c9d42583f346a95b5ec1bc572e0cb8e81f3bf9dfe1de0ef1e6 ``` for git 3b21f97

Raybuntu: We have to use full git sha in PKG_VERSION here.

projects/Amlogic/patches/linux/linux-001-PR94.patch
```diff
@@ -0,0 +1,92 @@
...
+ 	default:
+diff --git a/drivers/amlogic/display/osd/osd_hw.c b/drivers/amlogic/display/osd/osd_hw.c
+index 1040a57bafb4..1e70ce8577d6 100644
+--- a/drivers/amlogic/display/osd/osd_hw.c
++++ b/drivers/amlogic/display/osd/osd_hw.c
+@@ -2296,7 +2296,6 @@ void osd_pan_display_hw(u32 index, unsigned int xoffset, unsigned int yoffset)
+ 		osd_hw.pandata[index].y_start += diff_y;
+ 		osd_hw.pandata[index].y_end   += diff_y;
+ 		add_to_update_list(index, DISP_GEOMETRY);
+-		osd_wait_vsync_hw();
```
Raybuntu: @kszaq There are some reports that removing this line makes Netflix SW decodes stutter on S912. Any idea?

projects/Amlogic/packages/opengl-meson-t82x/package.mk
afl1: I did testing on KVIM2. I see lot of frame skips for  4k high bits testing videos in OSD debuging. Previous version runs skip free. I prefer performance for HW decoding over SW decoding improvement.

projects/Amlogic/patches/linux/linux-001-PR94.patch
kszaq: With the whole kernel patch applied and userspace libs updated to 3b21f97 I have not noticed stuttering.

kszaq:
Closing as too many issues with S912.

Raybuntu:
@kszaq You can leave it open. I'd still like to consider this after our final release. Up to you.

arthur-liberman:
@kszaq The PR will need to be updated, as I just pushed a fix for the issue with possibly setting values according to previous property data.
https://github.com/CoreELEC/linux-amlogic/commit/6f041639e769fb6d58d022bfa28e84f90ae4078a

wrxtasy:
Yes taking out Patch #1 from linux-001-PR94.patch is all that is needed now to get S912 SW decoding working again with Netflix after @arthur-liberman Kernel changes.

PR #164, opened 2019-02-15
GLES S912 - Correct version numbering and update to Android v3.2

wrxtasy:
Forgotten where the binaries were from, I think maybe Freaktab.

Tested with libyhbris and r5p1 UTGARD gpu-aml
Open GLES 3.2 is definitely faster with Android S912 due to decent ARM GPU tech.
S912 LE GUI may be a bit smoother.

This is a dirty patch...
@kszaq might want to update his t82x repo after testing.


kszaq:
Hey @wrxtasy. I updated my repo from latest VIM2 Nougat, I am not sure whether it's the same libs as from you: https://github.com/kszaq/opengl-meson-t82x/commits

It'd be also worth updating libhybris to support GLES3 for the future: https://github.com/kszaq/libhybris/commits/master

wrxtasy:
Hi @kszaq 

> I updated my repo from latest VIM2 Nougat, I am not sure whether it's the same libs as from you: https://github.com/kszaq/opengl-meson-t82x/commits

Your updated t82x are still v3.0 GLES3 my dirty patch above was a bump to v3.2

> It'd be also worth updating libhybris to support GLES3 for the future:

When used with you updated t82x GLES, Kodi compiling will error out with from memory a GLES3 includes Glint (may have been GLsync) error in ....
https://github.com/kszaq/libhybris/blob/master/hybris/include/GLES3/gl3.h#L1042

kszaq:
Try to use c21320b3c9bd043d05c95550a72b47f43016cfec for libhybris, the headers needed to be updated.

kszaq:
Note that Kodi will still link against GLESv2 - using updated libhybris from my repo will only enable using UNPACK_ROW_LENGTH that should result in a bit improved rendering with software decoding. Nothing more as Kodi does not offer any GLESv3-specific code.

wrxtasy:
OK is this to be used with r5p1 UTGARD gpu-aml or r7p0 ?

And GLES rendering improvement is better than nothing.

kszaq:
Libhybris is not used for Utgard, so you can use either.

If you want to use r7p0 Utgard, you need to apply a kernel patch for proper vsync and at the same time update gralloc for S912 to keep vsync working.

wrxtasy:
Good news is Kodi is compiling now without errors when using https://github.com/kszaq/libhybris/commit/c21320b3c9bd043d05c95550a72b47f43016cfec

kszaq:
@wrxtasy It looks that the library after patching matches this revision:
https://github.com/khadas/android_vendor_amlogic_common_gpu/blob/b1be27224570e0d6ecf16e0bb9af96a351270c45/lib/t82x_ion/libGLES_mali_default_8a_32-n-r16p0.so
And if you look at the commit history the version from my repo is a bit more recent:
https://github.com/khadas/android_hardware_arm_gpu/commits/Nougat/lib/t82x_ion/libGLES_mali_default_8a_32-n-r16p0.so

It looks like Amlogic wanted to pass CTS test by changing from 3.2 to 3.0, I am wondering why they did not change from 3.0 to 3.2 in build.prop...

I am not using S912 device so it's hard for me to tell which version of the library gives best results - it is quite possible that libhybris may work better with one of the previous versions.

wrxtasy:
@kszaq 
Cheers mate. 
Yes AML are being forced to get all their ducks in a row for CTS testing it seems.

I will have a brief play around again with your t82x GLES repo.

boot2k3:
Compared both 3.2 and 3.0 many times...
With 3.2 for me GUI looks smoother on S912

kszaq:
I don't think that version number returned by library makes a difference, especially since Kodi uses GLES2.0. What makes a difference is how the library works with libhybris in this particular usage, i.e. emulating fbdev. It's entirely possible that using an older lib gives smoother GUI.

kszaq:
It's also possible that gralloc in this patch was optimized in some way by device vendor that it comes from.

PR #166, opened 2019-02-17
Kodi upstream patches for Subs-video frame stutter & 4th refresh switching option

wrxtasy:
Memphiz patch adds a 4th refresh switching setting "On Start" so Kodi refresh switches on video playback start only. Helps users with slow switching TV's and AVR's and un necessary refresh switching when watching same fps content. Great for watching Netflix TV series. 

Downside - the Kodi GUI will stay at that refresh rate. On my S912 testing it's hardly noticeable, even at 24p.

Kszaq patch - fixes the very first Overlay render -->> video frame upset that occurs when you first activate Subtitles. It's blocked in upstream Kodi at the moment due to a 6+ months old lrusak PR that I doubt will make it into Kodi Leia.

These can always be removed if PR's merged in upstream Kodi.


kszaq:
Only for the note: the patch set you merged improved subtitle rendering for all subtitles (not only the first one) and also software decoding performance on S912. 👍🏻

wrxtasy:
Thanx for the info.
The commit above has been changed by already by Adam to the one currently in https://github.com/xbmc/xbmc/pull/15518

Which one should we be using ?

kszaq:
IMO you'll benefit most from using all 3 commits. #15518 was modified to include only BGRA textures support and doesn't patch using UNPACK_ROW_LENGTH.

PR #168, opened 2019-02-20
eventlircd: add abox bluetooth remote and tweak Mi Box remotes

wrxtasy:
Update Bluetooth remotes for Kodi Leia usage


wrxtasy:
Good to go after I removed some leftover junk.

The cheap ABOX BT remote I'm talking about is the same as this bluetooth one:
https://www.youtube.com/watch?v=H97K5RsjERI

It' has lightning quite key repeat response and scrolls thru long file lists like a rocket !

PR #171, opened 2019-03-03
Subtitle_fix

boot2k3:
Modified https://github.com/CoreELEC/CoreELEC/pull/163 patch with fix for SW S912 decoding

Please test on S905 devices too
If any regression, than it possible to change PR for S912 only


wrxtasy:
Pls. Don't merge.

I tested this exact setup a while back and still had external subtitles stuttering on S912.
I will re-test later on tonight to see if anything has changed.

adamg88:
needs thoroughly testing on gxbb/gxl before merging, last time we prematurely bumped these packages 6 months ago it caused big problems for a huge group of users

boot2k3:
> @boot2k3 confirming the recent test .tar you posted has this PR included ?
Yes, .tar I sent for testing include this PR + OpenGL ES 3.2 PR 164. But I tested without PR 164 too, the same result related to subtitles.
Need to test on S905 devices to decide if we need to include them or crop this patch to use it only for S912.

adamg88:
Are you referring to this one @kszaq ?

https://github.com/CoreELEC/linux-amlogic/commit/129d3030e5b6328a063cecba33d23a3f48b852f0

kszaq:
Yes. I don't understand the `if` in this commit but it looks like it returns `true` in my case and that means double wait for vsync. I think that if you want to use r7p0 for S905, you should make kernel work like Hardkernel's implementation, i.e. drop https://github.com/CoreELEC/linux-amlogic/commit/129d3030e5b6328a063cecba33d23a3f48b852f0 and fix S912 Mali libs.

My approach to fixing S912 was patching gralloc: http://ix.io/1CDf
But it looks like it doesn't work for everybody.

kszaq:
In other words: for S912 try to drop the kernel commit and use this gralloc: https://github.com/kszaq/opengl-meson-t82x/raw/alternate_vsync/system/lib/hw/gralloc.amlogic.so

I'm not saying that this works in all cases but I never had issues with it also for watching Netflix.

boot2k3:
> In other words: for S912 try to drop the kernel commit and use this gralloc: https://github.com/kszaq/opengl-meson-t82x/raw/alternate_vsync/system/lib/hw/gralloc.amlogic.so
> 
> I'm not saying that this works in all cases but I never had issues with it also for watching Netflix.

Thank you for testing
Yep, looks like `if` part need to be used for gxm only.

As for gralloc... I tried different ways and sources and tested this gralloc. Without this change on S912 either subtitles work without issue or SW part works without issue. For both parts we need this workaround. 
Any better solutions for S912 are welcome.

PR #188, opened 2019-03-26
Revert "projects/Amlogic: enable SSH by default"

kszaq:
"ssh" in cmdline forces SSH server to be on, user should be able to disable it.

This reverts commit 45380490ccd6138a67a33cf9cb1238da7cf10456.


adamg88:
Agreed, was just added for convenience initially.

PR #203, opened 2019-04-12
Lib input handler

Portisch:
Just tested it on linux 3.14 and it is working.


Raybuntu:
Thanks for the great work!

PR #207, opened 2019-04-22
kodi: add PR15941 to fix GLES HQ scalers

kszaq:
This PR adds https://github.com/xbmc/xbmc/pull/15941 to Kodi as the patch is not included in 18.2 final.

Without this patch all "HQ" SW scalers blur instead of sharpen (at least that's how it looks like) - screenshots available in Kodi PR.. I'm not sure if S905/S912 are powerful enough to use HQ scalers but Odroid-N2 should certainly benefit from this patch.

The patch can be dropped if/when merged to Kodi.


adamg88:
Thanks again @kszaq 👍

PR #211, opened 2019-06-27
kodi: fix PVRManager component dependencies

Koumes:
Kodi occasionally crashes when switching TV channels.

Details: https://github.com/xbmc/xbmc/issues/16303


arthur-liberman:
Interesting. I don't think I personally had Kodi crash in this way, but I don't switch TV channels that often.
Have you confirmed that this patch is applied cleanly, that build passes, that it fixes the problem and doesn't introduce new issues?

afl1:
I also had this occasional  kodi crashes at channel zapping. I already push this fix as we used  different source for kernel 3.14 and kernel 4.9 versions.

Koumes:
@arthur-liberman: It was built successfully (PROJECT=Amlogic SYSTEM=S912 ARCH=arm) and didn't see any added issues.

@afl1: Thank you very much.

spycat88:
Closing as merged separately.

PR #214, opened 2019-07-27
Add RTL8822BU support

jarda-wien:
Add support for RTL8822BU based WIFI dongles


jarda-wien:
the git log is dirty on this and needs a cleanup

PR #221, opened 2019-10-08
avahi: remove disallow-other-stacks=yes

tmm1:
This was originally added because SO_REUSEPORT was not available on older kernels

With this setting, avahi-daemon does not use either SO_REUSEADDR or SO_REUSEPORT, which means that other mDNS applications cannot access the multicast udp port 5353 at the same time. This includes mDNS clients which are simply trying to discover other services on the network. One such client exists in a PVR add-on I am writing for Kodi which currently fails in network discovery when run on CoreELEC.

See also grandcat/zeroconf#63

### DIFF
--- a/packages/network/avahi/package.mk
+++ b/packages/network/avahi/package.mk
@@ -57,8 +57,6 @@ pre_configure_target() {
 }
 
 post_makeinstall_target() {
-# for some reason avai can fail to start see: http://forums.gentoo.org/viewtopic-p-7322172.html#7322172
-  sed -e "s,^.*disallow-other-stacks=.*$,disallow-other-stacks=yes,g" -i $INSTALL/etc/avahi/avahi-daemon.conf
 # disable wide-area
   sed -e "s,^.*enable-wide-area=.*$,enable-wide-area=no,g" -i $INSTALL/etc/avahi/avahi-daemon.conf
 # publish-hinfo


spycat88:
> This was originally added because SO_REUSEPORT was not available on older kernels

A large proportion of our users are still running CE with the 3.14 Kernel, have you tested this change on this kernel?

tmm1:
> A large proportion of our users are still running CE with the 3.14 Kernel, have you tested this change on this kernel?

Okay, in that case a better fix would be to use this patch that debian/gentoo are using: https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/net-dns/avahi/files/avahi-0.6.31-so_reuseport-may-not-exist-in-running-kernel.patch?revision=1.1

arthur-liberman:
@adamg88 We can add a condition where `disallow-other-stacks=yes` will only be applied in case of 3.14 kernel. Wouldn't that be an acceptable solution?

spycat88:
@tmm1 I'm cool with that patch, can you amend your PR and we can get it merged.

@arthur-liberman I would like to keep things like that to a minimal if we can help it.

tmm1:
Is there any example I can look at for how to download and apply the patch before build?

spycat88:
@tmm1 https://github.com/CoreELEC/CoreELEC/blob/coreelec-9.2/packages/network/avahi/patches/avahi-0001_fix-autoconf.patch

Just create a new patch file in the same directory.

tmm1:
Thank you. I have imported the patch into the tree.

arthur-liberman:
Did you make sure that this patch applies cleanly? `PROJECT=Amlogic-ng ARCH=arm scripts/unpack avahi`
Also, you will need to rebase your branch and drop the previous commit.

tmm1:
> `PROJECT=Amlogic-ng ARCH=arm scripts/unpack avahi`

Thanks. I tried this, and it looks like avahi 0.7 already contains this patch. So no changes to the source are required.

> Also, you will need to rebase your branch and drop the previous commit.

I have rebased my branch. The original commit dropping the configuration option is still required, and the new configuration works fine on older kernels with avahi 0.7

tmm1:
Ping?

arthur-liberman (approved):

arthur-liberman:
Sorry for the delay, and thanks for the contribution

PR #237, opened 2020-04-16
ceemmc: bump to 63497b50

francis2400:


arthur-liberman:
What is this for?

Portisch:
I guess it was a accident. Close PR as it would downgrade the tool.

PR #240, opened 2020-05-08
librespot: enable running on CoreELEC with CONFIG_CP15_BARRIER_EMULATION

vpeter4:
the other option is to disable emulation in Linux kernel
(I don't understand what is this even for ...)

Edit: Rust needs kernel option and without it doesn't even run.


arthur-liberman (approved):

Portisch:
Disable of Kernel config flag CONFIG_CP15_BARRIER_EMULATION does not solve the issue.
So reopened the PR and merge it.

PR #241, opened 2020-05-18
Fix compilation error for squashfs

jedi7:
Fixes following compilation error
read_fs.o:(.bss+0x0): multiple definition of `fwriter_buffer';
read_fs.o:(.bss+0x8): multiple definition of `bwriter_buffer'; mksquashfs.o:

It happen on new gcc which force -fno-common


jedi7:
Nope.
Base is from https://bugs.gentoo.org/706456
The toolchain was generated automatically with version: gcc (GCC) 10.1.0
Not sure if it is good version for embedded dev, but I just run make.

Portisch:
Thank you for the hint of this patch!

PR #242, opened 2020-05-19
Fix build against gcc 10

jedi7:
Fixes compilation problem:

scripts/dtc/dtc-parser.tab.o:(.bss+0x50): multiple definition of `yylloc'; scripts/dtc/dtc-lexer.lex.o:(.bss+0x0)

(there will be more packages to fix)

### DIFF
--- a/packages/compress/cpio/patches/fix_build_against_gcc_10.patch
+++ b/packages/compress/cpio/patches/fix_build_against_gcc_10.patch
@@ -1,0 +1,14 @@
+diff --git a/src/global.c b/src/global.c
+index 57e505a..5106271 100644
+--- a/src/global.c
++++ b/src/global.c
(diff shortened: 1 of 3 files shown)


Portisch:
Please add the other package fixes as seperate commits to this PR.

jedi7:
fixes complete, compilable now with gcc 10.
btw: is there way to change the version of toolchain gcc?

Portisch:
Currently CE use 8.3.0: https://github.com/CoreELEC/CoreELEC/blob/coreelec-9.2/packages/lang/gcc/package.mk
LE updated to 9.3.0 in master branch: https://github.com/LibreELEC/LibreELEC.tv/commits/master/packages/lang/gcc/package.mk

PR #246, opened 2020-05-30
Fix compilation errors on GCC 10 - take two

adrian-007:
There's already a pull request for that issue, I just picked relevant patches from upstream project and added a references to them in patches.

I also bumped binutils version and patched build of p7zip as that fixed building on Fedora 32.

PR #248, opened 2020-06-03
nghttp2 memleak fix + Gcc10 fixes

svde:
See this report: https://discourse.coreelec.org/t/odroid-c2-playing-from-davs-https-share-broken-since-9-0-2
based on: https://github.com/CoreELEC/CoreELEC/pull/246 in order to build it on fc32 with gcc10


cdu13a:
I appreciate the pull request. However to make the pull request easier for the developers to evaluate this should be two separate pull requests. One for anything related to the nghttp2 package bump, and one for the gcc10 fixes.

There should also be a clearer explanation for the package bumps as well.

svde:
OK. I guess I need to split it all out into super small pieces. For each tiny bit a new branch.

cdu13a:
It's not so much just about small pieces as it is that one pull request should do one thing, and the title and description should be clear about what that is, and why the pull request is being made. Additional notes such as how and on what devices you tested are also a good idea when relevant. 

The individual commits that make up a pull request should also each only do one thing, and have a clear title and description that uses the convention that the rest of the commits in the repo use.

I didn't notice it this morning probably because I hadn't finished my morning coffee yet when I initially viewed your pull request, but including commits in your pull request that are duplicates of commits found in another open pull request will pretty much make it so that your pull request will not be able to be merged. If your pull request requires an other open pull request to be merged first this should be indicated in the description.

svde:
It built on fc32/x86_64, and I have the resulting image running on my Odroid C2. The memory leak which caused the odroid to crash when streaming from http/2 webservers is gone.

The gcc10 fixes were needed to get it to build, as fc32 is on that compiler version.

PR #249, opened 2020-06-03
nghttp2: bump to 1.41.0

svde:
Update nghttp2 to 1.41.0 to fix issues with accessing content from a davs or https directory on a server using HTTP/2.0. 1.41.0 also fixes CVE-2020-11080

See this report: https://discourse.coreelec.org/t/odroid-c2-playing-from-davs-https-share-broken-since-9-0-2


svde:
Just saw that 1.41.0 is also released, also fixes CVE-2020-11080.

cdu13a:
This is better, though you only need "nghttp2: bump to 1.41.0" since having two bumps to the same package in one pull request just clutters things up, when just the one will do.

The linking to the form post is good for further context, however there still needs to be a description in addition to the link.

So if this is indeed what the pull request is for a description such as the following would be required.

`Update nghttp2 to 1.41.0 to fix issues with accessing content from a davs or https directory on a server using HTTP/2.0`

That way it is clear to the person reviewing the pull request what is being done and why.

Now is the bump to nghttp2, everything that was needed to fix the issue this is in relation to?

svde:
> Now is the bump to nghttp2, everything that was needed to fix the issue this is in relation to?

I can't test it without the gcc10 fixes, as my build system needs that to be fixed.
In the other PR I also took along updates for the other web packages, which aren't in this PR.

cdu13a:
The other web packages were not required as part of the solution to your problem?

As for dealing with other commits that are specific to your build enviroment, that you require to build, but are not needed as part of the pull request. You can create a branch and have all your commits you need to test and develop with in it, then when it comes time to make a pull request you can create a new branch based on the current coreelec-9.2 branch and then include just the commits that are required to be part of the pull request in it.

svde:
> The other web packages were not required as part of the solution to your problem?

Only tested with the other web package updates.

> As for dealing with other commits that are specific to your build enviroment, that you require to build, but are not needed as part of the pull request. You can create a branch and have all your commits you need to test and develop with in it, then when it comes time to make a pull request you can create a new branch based on the current coreelec-9.2 branch and then include just the commits that are required to be part of the pull request in it.

Yes, starting to figure this out, in the midst of a learning curve.

I guess I could also not share what I've fixed. I was the only one on discourse who reported the issue and there wasn't much interest in it either. Just compile & run for myself.

cdu13a:
To fix up the two bump commits so that there is only the 1.41.0 one

`git reset --soft HEAD^`
`git commit --amend`

This will uncommit the newer commit while keeping it's changes and add it's changes to the first commit, then you just change the commit message and push the branch again.

after that I will test to verify everything is ok, and if at that point if everything is good, I will merge your pull request.

cdu13a:
Yes it is a bit of a learning curve, that is not as well documented as it could be. However you are getting there, and the pull request is appreciated.

Unfortunately we do have to insist on some things being done a certain way, since things can get a little chaotic to manage otherwise.

svde:
So... from two to four commits.

cdu13a:
ok, to get back down to the 1 commit that is needed. 

`git reset --hard HEAD^`
Will get rid of the merge commit that ended up multiplying the number of commits. after doing so use `git log` to verify that you no longer have the additional commits from the merge commit.

now `git push --force` can be used to update the pull request.

cdu13a:
I can guarantee you that everyone of the CoreELEC developers went through a similar learning curve when we first started. You eventually figure it out and get comfortable with git and how everything works.

It's just a matter of practice, and being patient while you learn.

cdu13a:
ok, just did a test build with your pull request.

Since this package downloads the nghttp2-1.41.0.tar.xz source archive instead of the nghttp2-1.41.0.tar.gz source archive. The PKG_SHA256 you used is wrong and should actually be

`PKG_SHA256="abc25b8dc601f5b3fefe084ce50fcbdc63e3385621bee0cbfa7b57f9ec3e67c2"`
so that it matches the files being downloaded.

Make the change to the package file then use the following 3 commands to update the pull request.

`git add packages/web/nghttp2/package.mk`
`git commit --amend`
`git push --force`

I've uploaded both the amlogic and amlogic-ng update file for the test build so that you can verify that everything is working as expected https://test.coreelec.org/cdu13a/public.builds/

svde:
I think I'm making a mess. Perhaps try again another day.

PR #254, opened 2020-07-05
bluez: update to bluez-5.54 / ethtool: update to ethtool-5.4

SimplyRyan:


Portisch:
I tried to compile this package bumps but the package patches needs to be rebased.
As I have seen LE/master have already bumped the packages and they should be included in coreelec-10.0.

PR #255, opened 2020-07-11
Amlogic-ng: Support 4K GUI scaling on boot.

Raybuntu:
Corresponding CE PR for https://github.com/CoreELEC/xbmc/pull/1
I think a 2 linux reverts are also required. I will add the kodi bump and the linux bump to this PR when the other PR's are merged.

### DIFF
--- a/projects/Amlogic-ng/initramfs/platform_init
+++ b/projects/Amlogic-ng/initramfs/platform_init
@@ -25,6 +25,9 @@ for arg in $(cat /proc/cmdline); do
     frac_rate_policy=*)
       frac_rate_policy=${arg#*=}
       ;;
+    native_4k_gui=*)
+      native_4k_gui=${arg#*=}
+      ;;
   esac
 done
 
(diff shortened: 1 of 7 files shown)


projects/Amlogic-ng/initramfs/platform_init
```diff
@@ -51,9 +54,15 @@ case $display_mode in
   480*)            X=720  Y=480  ;;
   576*)            X=720  Y=576  ;;
   720p*)           X=1280 Y=720  ;;
+  2160p*)          X=3840 Y=2160 ;;
   *)               X=1920 Y=1080 ;;
 esac
 
+if [ "$native_4k_gui" == "0" ]; then
```
arthur-liberman: I think you should only fallback to 1080p if the current X or Y correspond with a higher resolution.

Raybuntu: Ohh right. Bug

PR #258, opened 2020-08-08
Bluetooh gamepads 

Bermarck:
I tried emuelec on TX3MINI with SOC s905w
Emuelec runs very well. Great job. 
Pairing mode is easier than before, but my Bluetooth couldn't be configured  hotkey on retroarch although on emulestation menu seems to be configured but when I played games I couldn't get out. I was looking for the way on retroarch but buttons a, x, y, b don't work just I can move. On emulestation am main menu it works w very good. Is there anyway to configure Bluetooth gamepads hotkey? 
I tried with I pega 9128 don't work 
Ps4 gamepad work with cable 
Other generic gamepad works on emulestation but buttons select and start don't work when I want to play games. 
You are doing a good work. Thanks a lot.


shantigilbert:
Wrong repo https://github.com/EmuELEC/EmuELEC

PR #259, opened 2020-08-17
Use fakeroot with TCP.

V10lator:
This fixes building on WSL (see https://github.com/microsoft/WSL/issues/2465 , for example).


Portisch:
Version 1.24 is not available. I tried with 1.25.3 and build failed.
If still required please check if again and we can re-open the pull request, thank you!

PR #264, opened 2020-10-19
Amlogic-ng: enable Analog TV tuner STK1160

Portisch:
I am not sure if this should be added. It enable Analog TV tuner support used for STK1160.

Confirm by user with compatible hardware is still missing.


Portisch:
The driver got tested and confirmed working in Amlogic-ng 4.9 kernel.

PR #267, opened 2020-11-19
Unbreaking libhybris package for aarch64 ARCH and GCC 8

dmikushin:
This patch bumps libhybris to address the fixes described in https://github.com/kszaq/libhybris/pull/1

### DIFF
--- a/packages/graphics/libhybris/package.mk
+++ b/packages/graphics/libhybris/package.mk
@@ -2,17 +2,22 @@
 # Copyright (C) 2018-present Team CoreELEC (https://coreelec.org)
 
 PKG_NAME="libhybris"
-PKG_VERSION="c21320b3c9bd043d05c95550a72b47f43016cfec"
-PKG_SHA256="c71d8dc18b4784ad80c120869ec21b34fcdd1503be738f3ea0f51b9489001c60"
+PKG_VERSION="234a0480c887c29f8d6a895ac86a350e4e0fe05a"
+PKG_SHA256="18551bee3d660bca744324b3fb5477bec1796f3b3fb8b15808fc43e248dd0ffb"
 PKG_LICENSE="GPL"
-PKG_SITE="https://github.com/kszaq/libhybris"
-PKG_URL="https://github.com/kszaq/libhybris/archive/$PKG_VERSION.tar.gz"
+PKG_SITE="https://github.com/dmikushin/libhybris"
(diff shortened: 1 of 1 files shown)


arthur-liberman:
This is a nice workaround for having libhybris built while having the rest of the image run in 64-bit mode, but why should we bump the package? Did you change anything in it?
Also, have you tested the resulting image? On which device, and using which kernel?
libhybris is only used in S912 boxes, and only with the 3.14 kernel (non-NG).

dmikushin:
Hi, I just wanted to get the build process to complete successfully. I'm not aware whether libhybris is used on not, but without fixing it, I cannot build with Docker image for the **Amlogic-ng**. Hence, here is a patch.

dmikushin:
> This is a nice workaround for having libhybris built while having the rest of the image run in 64-bit mode

I think libhybris is also 64-bit mode, as `aarch64-libreelec-linux-gnueabi-gcc` is deployed without any special option. The libhybrid only used the notion of "arm" to distinguish from "x86". Hence, libhybrid and everything is 64-bit.

> why should we bump the package? Did you change anything in it?

I've fixed compiler errors. Otherwise, it's quite surprising that you could somehow compile it for yourself. I think you deploy a different (older?) toolchain, not the one offered by https://github.com/CoreELEC/coreelec-builder

packages/graphics/libhybris/package.mk
```diff
@@ -2,17 +2,22 @@
...
 PKG_NAME="libhybris"
-PKG_VERSION="c21320b3c9bd043d05c95550a72b47f43016cfec"
-PKG_SHA256="c71d8dc18b4784ad80c120869ec21b34fcdd1503be738f3ea0f51b9489001c60"
+PKG_VERSION="234a0480c887c29f8d6a895ac86a350e4e0fe05a"
+PKG_SHA256="18551bee3d660bca744324b3fb5477bec1796f3b3fb8b15808fc43e248dd0ffb"
 PKG_LICENSE="GPL"
-PKG_SITE="https://github.com/kszaq/libhybris"
-PKG_URL="https://github.com/kszaq/libhybris/archive/$PKG_VERSION.tar.gz"
+PKG_SITE="https://github.com/dmikushin/libhybris"
+PKG_URL="https://github.com/dmikushin/libhybris/archive/$PKG_VERSION.tar.gz"
```
arthur-liberman, line 9: What is this change for?

Raybuntu:
We don't support 64bit userspace. We also won't support libhybris in the future

PR #268, opened 2020-11-30
Enable lock LED for TBS 5520SE

smp79:
This will make the green "signal lock" LED functional on TBS 5520SE DVB tuner.


Raybuntu:
I can't test this but if it still builds I have no objections.

Portisch:
This is to much device specific as this driver is used also for other hardware.
We can not be sure it does not affect other tuner hardware.

PR #269, opened 2020-12-27
Set correct branch for linux-amlogic

V10lator:


Portisch:
Thank you for your support!

PR #271, opened 2021-01-04
Add cryptsetup and libdevmapper packages

doozan:
This patch adds the cryptsetup binary, allowing users to create and mount encrypted devices.

Cryptsetup depends on libdevmapper, which is part of the lvm2 (not currently included in LE). The library is only 366k but the lvm2 binaries are 4M so instead of including them I've created a libdevmapper package that installs just the library. I'm open to suggestions if there's a better way of doing this.

In total, this adds about 870k (uncompressed) to the system:

```
./usr/lib:
lrwxrwxrwx    1 1000     1000          23 Jan  1 12:00 libcryptsetup.so -> libcryptsetup.so.12.6.0
lrwxrwxrwx    1 1000     1000          23 Jan  1 12:00 libcryptsetup.so.12 -> libcryptsetup.so.12.6.0
-rwxr-xr-x    1 1000     1000      378.5K Jan  1 12:00 libcryptsetup.so.12.6.0
lrwxrwxrwx    1 1000     1000          20 Jan  1 12:00 libdevmapper.so -> libdevmapper.so.1.02
-r-xr-xr-x    1 1000     1000      366.8K Jan  1 12:00 libdevmapper.so.1.02

./usr/lib/tmpfiles.d:
-rw-r--r--    1 1000     1000          35 Jan  1 12:00 cryptsetup.conf

./usr/sbin:
-rwxr-xr-x    1 1000     1000       93.4K Jan  1 12:00 cryptsetup
```


Portisch:
Can you please fixup `packages/sysutils/cryptsetup/package.mk`. There are quotes missing, thank you!

Raybuntu:
I would like to see this in CE19 too. +1 for this PR once the formatting issues are resolved.

Portisch:
Yes, CE19 would be better. @doozan please rebase it on `coreelec-19` branch. 9.2 is going to be EOL soon.

vpeter4:
Why is _O_DIRECT_ disabled in package.mk? Can't this be removed (or enabled) and then patch file is not needed.

PR #212, opened 2019-07-13
Add modules needed for LUKS

doozan:
This adds the modules needed to use LUKS.

When `CONFIG_ARM64_CRYPTO` is defined, the macro `module_cpu_feature_match()` is used to check that the CPU supports the crypto extension.  Unfortunately, that macro won't compile and gives the error:

```
In file included from arch/arm64/crypto/aes-glue.c:16:0:
include/linux/module.h:87:32: error: storage size of '__mod_cpu_device_table' isn't known
 extern const struct gtype##_id __mod_##gtype##_table  \
                                ^
include/linux/module.h:145:3: note: in expansion of macro 'MODULE_GENERIC_TABLE'
   MODULE_GENERIC_TABLE(type##_device, name)
   ^~~~~~~~~~~~~~~~~~~~
include/linux/cpufeature.h:49:1: note: in expansion of macro 'MODULE_DEVICE_TABLE'
 MODULE_DEVICE_TABLE(cpu, cpu_feature_match_ ## x);  \
 ^~~~~~~~~~~~~~~~~~~
arch/arm64/crypto/aes-glue.c:442:1: note: in expansion of macro 'module_cpu_feature_match'
 module_cpu_feature_match(AES, aes_init);
 ^~~~~~~~~~~~~~~~~~~~~~~~
make[1]: *** [/src/CoreELEC-coreelec-9.0/build.CoreELEC-Amlogic.arm-9.0-devel/linux-f1bb00821950b0f5cab1b8a820c697ba9fbd3183/arch/arm64/crypto/Makefile:38: arch/arm64/crypto/aes-glue-ce.o] Error 1
```

I've included a patch to remove the macro. Since we're targeting a specific device and it will have the expected extensions it should be safe, but I'm open to suggestions on how to better fix this.

The patch is important, because without the arm crypto, it's pretty slow.

Before:

```
CoreELEC:~ # cryptsetup benchmark
# Tests are approximate using memory only (no storage IO).
PBKDF2-sha1       281875 iterations per second for 256-bit key
PBKDF2-sha256     569878 iterations per second for 256-bit key
PBKDF2-sha512     238312 iterations per second for 256-bit key
PBKDF2-ripemd160  152409 iterations per second for 256-bit key
PBKDF2-whirlpool     N/A
argon2i       4 iterations, 169514 memory, 4 parallel threads (CPUs) for 256-bit key (requested 2000 ms time)
[...]

PR #272, opened 2021-01-05
Add cryptsetup package

doozan:
This is #271, but on 19.  As a bonus, 19 already has a libdevmapper package, so no need to duplicate that.


vpeter4:
Header missing in cryptsetup package.mk.

doozan:
Header added

Raybuntu:
> Header added

Where?

doozan:
On my local machine, apparently :)  It's there now.

Portisch:
How to use this crypt package? Would there not be needed something in busybox/init script to decrypt the drive? How to encrypt the drive? Would be `COREELEC` partition also be encrypted?

doozan:
The binary just gives makes it possible to create/open/close encrypted volumes, and it's probably most useful for users that want to attach encrypted media to their device. Here's a [short intro](https://linuxconfig.org/basic-guide-to-encrypting-linux-partitions-with-luks) on using it.

It's possible to encrypt COREELEC if you add a custom mount-storage.sh init script to handle the unlocking. See https://github.com/CoreELEC/CoreELEC/pull/212#issuecomment-511240178 for a very rough example script that starts sshd and waits to for the user to connect and unlock the volume.

Portisch:
Closed by https://github.com/CoreELEC/CoreELEC/commit/3ee93100c8a862666484c42f17c366df3105d475
Thank you for your support!

PR #273, opened 2021-01-17
add missing package to docker image

bendavid:
Building following instructions in tools/docker/README.md fails without this.


arthur-liberman:
Is this required only for CE? Because if LE can also use it, it's better to just directly upstream it to them.

bendavid:
Probably should go upstream then yes.

arthur-liberman:
Ok, then please close this PR and open a PR to LE. Once they merge it, we'll pull the change from them.

PR #276, opened 2021-03-14
aml-vnc: bump to f87a8f4

kszaq:
I recently needed to support family over VNC and it was a little bit painful - here's an improvement for a better experience. No mouse support yet, unfortunately.


Raybuntu:
Hi @kszaq. Nice to hear again from you. I hope you are healthy. Let me get that merged so it can be tested in nightly.

kszaq:
HI @Raybuntu. It's been a long time since I retired from any *ELEC tinkering but always happy to contribute if I consider something useful.

arthur-liberman:
Hi @kszaq, I was wondering, do you know why HDR10/HLG VP9 streams do not trigger HDR, like from YouTube, or even local files?
I have always assumed, maybe incorrectly, that it has to be something that's missing from here: https://github.com/CoreELEC/xbmc/blob/aml-4.9-19.1/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp#L923
As you were the one who added VP9 support originally for AML in LE, I thought maybe you know.

kszaq:
I wasn't aware of that issue. The code added to Kodi was adapted from amplayer at that time, if you have any recent source code from Amlogic you may have a look if there was any HDR-related update.

It looks like the code did not change for 4.9: https://github.com/vitmod/libplayer-4.9/blob/master/amplayer/player/player_hwdec.c#L1418

arthur-liberman:
Yeah I remember digging around and finding that original code that you used.
But if the code did not change, then I'm not sure what we are missing. It's possible that the player still doesn't support VP9 HDR.
I see that the last commit in that branch was made in 2017. :(

arthur-liberman:
Looks like nothing changed in 5 years. Maybe the issue is somewhere else.
https://github.com/numbqq/libplayer/blob/master/amplayer/player/player_hwdec.c#L1375

PR #285, opened 2021-04-11
inputstream.adaptive: bump to 2.6.13-Matrix

matthuisman:
some pretty big important fixes since 2.6.7 have come in that fix quite a few add-ons


Portisch:
Wrong branch and package is maintained by LE.

Issue #10, opened 2018-04-16
PVR/VNSI issues


TheChief79:
As i already wrote in LE-forum, me and friend having problems with LiveTV in 8.90.0. No audio and video stops after few seconds. We have both H96 Pro + with 912 SOC and use VDR/vnsiserver/VNSI Addon.

Logfiles attached:

[log-2018-04-16-07.17.15.zip](https://github.com/CoreELEC/CoreELEC/files/1917247/log-2018-04-16-07.17.15.zip)

Just opend that bug report again for better visibility and to track the status.

Thanks for your work.

adamg88:
Have you tried TVH to see if it is just an add-on related issue rather than a build issue?

TheChief79:
No, but i can try if i find some time. But i think it could be addon related, because dvbviewer addon seems to work for my friend. On the other hand, everything worked with vnsi until the new release.

TheChief79:
TVH is working, audio and video.

afl1:
I pushed fix for this issue. In next release it'll be fixed.
[https://github.com/CoreELEC/CoreELEC/commit/cc036516eef967cd992323e684f3b53bb027814f](url)

TheChief79:
Cool thanks. I would build by myself, but it takes too long :D

TheChief79:
Vnsi works with your fix but as i already told boot2k3, remote does not work die to latest kodi changes.

https://github.com/xbmc/xbmc/pull/13761

adamg88:
Please keep all comments here so other developers know what you are referring to as myself and others are not privy to your conversation with boot2k3.

TheChief79:
Ok, he just sent me a developer package with the fix from afl1:  https://github.com/CoreELEC/CoreELEC/commit/cc036516eef967cd992323e684f3b53bb027814f

With this fix vnsi works again.

PR #287, opened 2021-05-17
[ARM] Add suport for libwidevine.so >= 4.10.2252.0

Portisch:
@kszaq @wagnerch 
As it looks like LibreELEC will not merge it CoreELEC want to pick the PR:
https://github.com/LibreELEC/LibreELEC.tv/pull/5376

Please can you check the last four commits if I picked them correct?
I like more the way not using the LD_PRELOAD. It compiles fine but I can not test it.

Also we prefer to add a commit in XBMC directly instead adding a patch, especially without header, no author, no date, no description...

@wagnerch please check the patch if it's correct:
https://github.com/CoreELEC/CoreELEC/blob/e9531ed494b412330c262fe2d37c0cb10069d296/projects/Amlogic-ce/devices/Amlogic-ng/patches/kodi/kodi-temp-100.35-add-tcmalloc-support.patch
or provide a better one as otherwise I will pick this patch with current header I have copied from your last patch.

Thank you!

### DIFF
--- a/config/options
+++ b/config/options
@@ -36,8 +36,10 @@ TARGET_ARCH="${ARCH}"
 # include arm-mem package on arm
 if [ "${TARGET_ARCH}" = "arm" ]; then
   ARM_MEM_SUPPORT="yes"
+  TCMALLOC_SUPPORT="yes"
 else
   ARM_MEM_SUPPORT="no"
+  TCMALLOC_SUPPORT="no"
 fi
 
 # include helper functions
--- a/packages/devel/glibc/patches/glibc-add-support-for-SHT_RELR-sections.patch
+++ b/packages/devel/glibc/patches/glibc-add-support-for-SHT_RELR-sections.patch
@@ -1,0 +1,348 @@
+From 6676e967dba405ca31d57b63e096becd13d4a200 Mon Sep 17 00:00:00 2001
+From: Rahul Chaudhry <rahulchaudhry@chromium.org>
+Date: Thu, 15 Mar 2018 14:30:17 -0700
+Subject: [PATCH 4/8] sys-libs/glibc: add support for SHT_RELR sections.
+
+This patch adds experimental support for SHT_RELR sections, proposed
+here: https://groups.google.com/forum/#!topic/generic-abi/bX460iggiKg
+
+SHT_RELR sections are supported for arm, aarch64, and x86_64 targets.
+To enable them, pass '--experimental-use-relr' flag to gold.
+
+Definitions for the new ELF section type and dynamic array tags, as well
+as the encoding used in the new section are all under discussion and are
+subject to change. We plan to send the patch upstream after the gABI has
+been updated to include the new definitions.
+
+[Adrian: forward-ported to glibc 2.32]
+---
+ elf/do-rel.h                 | 41 ++++++++++++++++++++++++++++++++++--
+ elf/dynamic-link.h           | 15 +++++++++++++
+ elf/elf.h                    | 15 +++++++++++--
+ elf/get-dynamic-info.h       |  7 ++++++
+ sysdeps/aarch64/dl-machine.h | 10 +++++++++
+ sysdeps/arm/dl-machine.h     | 10 +++++++++
+ sysdeps/i386/dl-machine.h    | 10 +++++++++
+ sysdeps/x86_64/dl-machine.h  | 10 +++++++++
+ 8 files changed, 114 insertions(+), 4 deletions(-)
+
+diff --git a/elf/do-rel.h b/elf/do-rel.h
+index 1d0a1f2c5d..25babef6e1 100644
+--- a/elf/do-rel.h
++++ b/elf/do-rel.h
+@@ -26,6 +26,12 @@
+ # define elf_machine_rel_relative	elf_machine_rela_relative
+ #endif
+ 
++#ifdef DO_RELR
++# define elf_dynamic_do_Rel		elf_dynamic_do_Relr
++# define Rel				Relr
++# define elf_machine_rel_relative	elf_machine_relr_relative
++#endif
++
+ #ifndef DO_ELF_MACHINE_REL_RELATIVE
+ # define DO_ELF_MACHINE_REL_RELATIVE(map, l_addr, relative) \
+   elf_machine_rel_relative (l_addr, relative,				      \
+@@ -46,12 +52,12 @@ elf_dynamic_do_Rel (struct link_map *map,
+   const ElfW(Rel) *r = (const void *) reladdr;
+   const ElfW(Rel) *end = (const void *) (reladdr + relsize);
+   ElfW(Addr) l_addr = map->l_addr;
+-# if defined ELF_MACHINE_IRELATIVE && !defined RTLD_BOOTSTRAP
++# if defined ELF_MACHINE_IRELATIVE && !defined RTLD_BOOTSTRAP && !defined DO_RELR
+   const ElfW(Rel) *r2 = NULL;
+   const ElfW(Rel) *end2 = NULL;
+ # endif
+ 
+-#if (!defined DO_RELA || !defined ELF_MACHINE_PLT_REL) && !defined RTLD_BOOTSTRAP
++#if (!defined DO_RELA || !defined ELF_MACHINE_PLT_REL) && !defined RTLD_BOOTSTRAP && !defined DO_RELR
+   /* We never bind lazily during ld.so bootstrap.  Unfortunately gcc is
+      not clever enough to see through all the function calls to realize
+      that.  */
+@@ -80,8 +86,10 @@ elf_dynamic_do_Rel (struct link_map *map,
+   else
+ #endif
+     {
++# if !defined DO_RELR
+       const ElfW(Sym) *const symtab =
+ 	(const void *) D_PTR (map, l_info[DT_SYMTAB]);
++# endif
+       const ElfW(Rel) *relative = r;
+       r += nrelative;
+ 
+@@ -108,9 +116,36 @@ elf_dynamic_do_Rel (struct link_map *map,
+ 	if (l_addr != 0 || ! map->l_info[VALIDX(DT_GNU_PRELINKED)])
+ # endif
+ #endif
++
++#ifdef DO_RELR
++	  {
++	    ElfW(Addr) base = 0;
++	    for (; relative < end; ++relative)
++	      {
++		ElfW(Relr) entry = *relative;
++		if ((entry&1) == 0)
++		  {
++		    elf_machine_relr_relative (l_addr, (void *) (l_addr + entry));
++		    base = entry + sizeof(ElfW(Addr));
++		    continue;
++		  }
++		ElfW(Addr) offset = base;
++		while (entry != 0)
++		  {
++		    entry >>= 1;
++		    if ((entry&1) != 0)
++		      elf_machine_relr_relative (l_addr, (void *) (l_addr + offset));
++		    offset += sizeof(ElfW(Addr));
++		  }
++		base += (8*sizeof(ElfW(Addr)) - 1) * sizeof(ElfW(Addr));
++	      }
++	  }
++#else
+ 	  for (; relative < r; ++relative)
+ 	    DO_ELF_MACHINE_REL_RELATIVE (map, l_addr, relative);
++#endif
+ 
++#if !defined DO_RELR
+ #ifdef RTLD_BOOTSTRAP
+       /* The dynamic linker always uses versioning.  */
+       assert (map->l_info[VERSYMIDX (DT_VERSYM)] != NULL);
+@@ -179,6 +214,7 @@ elf_dynamic_do_Rel (struct link_map *map,
+ 				 skip_ifunc);
+ # endif
+ 	}
++#endif
+ #endif
+     }
+ }
+@@ -189,3 +225,4 @@ elf_dynamic_do_Rel (struct link_map *map,
+ #undef elf_machine_rel_relative
+ #undef DO_ELF_MACHINE_REL_RELATIVE
+ #undef DO_RELA
++#undef DO_RELR
+diff --git a/elf/dynamic-link.h b/elf/dynamic-link.h
+index 6727233e1a..4345df9949 100644
+--- a/elf/dynamic-link.h
++++ b/elf/dynamic-link.h
+@@ -76,6 +76,11 @@ auto inline void __attribute__((always_inline))
+ elf_machine_rela_relative (ElfW(Addr) l_addr, const ElfW(Rela) *reloc,
+ 			   void *const reloc_addr);
+ # endif
++# if ! ELF_MACHINE_NO_RELR
++auto inline void __attribute__((always_inline))
++elf_machine_relr_relative (ElfW(Addr) l_addr,
++			    void *const reloc_addr);
++# endif
+ # if ELF_MACHINE_NO_RELA || defined ELF_MACHINE_PLT_REL
+ auto inline void __attribute__((always_inline))
+ elf_machine_lazy_rel (struct link_map *map,
+@@ -190,6 +195,15 @@ elf_machine_lazy_rel (struct link_map *map,
+ #  define ELF_DYNAMIC_DO_RELA(map, lazy, skip_ifunc) /* Nothing to do.  */
+ # endif
+ 
++# if ! ELF_MACHINE_NO_RELR
++#  define DO_RELR
++#  include "do-rel.h"
++#  define ELF_DYNAMIC_DO_RELR(map, lazy, skip_ifunc) \
++  _ELF_DYNAMIC_DO_RELOC (RELR, Relr, map, lazy, skip_ifunc, 1)
++# else
++#  define ELF_DYNAMIC_DO_RELR(map, lazy, skip_ifunc) /* Nothing to do.  */
++# endif
++
+ /* This can't just be an inline function because GCC is too dumb
+    to inline functions containing inlines themselves.  */
+ # define ELF_DYNAMIC_RELOCATE(map, lazy, consider_profile, skip_ifunc) \
+@@ -198,6 +212,7 @@ elf_machine_lazy_rel (struct link_map *map,
+ 					      (consider_profile));	      \
+     ELF_DYNAMIC_DO_REL ((map), edr_lazy, skip_ifunc);			      \
+     ELF_DYNAMIC_DO_RELA ((map), edr_lazy, skip_ifunc);			      \
++    ELF_DYNAMIC_DO_RELR ((map), edr_lazy, skip_ifunc);			      \
+   } while (0)
+ 
+ #endif
+diff --git a/elf/elf.h b/elf/elf.h
+index 197b557d15..5b6da8e8ae 100644
+--- a/elf/elf.h
++++ b/elf/elf.h
+@@ -446,7 +446,8 @@ typedef struct
+ #define SHT_PREINIT_ARRAY 16		/* Array of pre-constructors */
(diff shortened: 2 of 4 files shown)


kszaq:
@Portisch You can check if at runtime `ldd /usr/lib/kodi/kodi.bin` is linked to `tcmalloc_minimal`. If you can provide me with a test build, I can check today evening if the latest widevine blob works.

https://github.com/xbmc/inputstream.adaptive/issues/678#issuecomment-841667163

Keep in mind that introducing these patches has no known effects in short- or long-term run for Kodi. I am basically running "vanilla" Kodi with IA, so I haven't checked any other addons etc.

Portisch:
I currently do not have a device online so here my compile with including this PR:
[CoreELEC-Amlogic-ng.arm-19.2-Matrix_devel_20210517082302.tar](https://mega.nz/file/PShAAbRT#dZ6AMPazt-XKF2ZkTyLzlhio6OVETI2Omhi_9QbUeZE)

At least I find the `tcmalloc_minimal` in the `build` folder :wink:

If you have already CE-19 running just enter SSH:
```text
cd .update
megadl https://....
sync & reboot
```

kszaq:
glibc patch with more readable patch name (stripped ChromeOS prefix): 
https://github.com/LibreELEC/LibreELEC.tv/pull/5376/commits/20303ada83e66bee26364a1836cb4c82fea32804

wagnerch:
@Portisch It looks good.  It is the same patches that I applied to CoreELEC.  I haven't tested it as much as my Raspberry Pi, though.

https://github.com/CoreELEC/CoreELEC/compare/coreelec-19...wagnerch:widevine_support

Portisch:
Ok, thank you. Then I will finish up the PR to be ready to be merged as soon someone can confirm it's working.

andy5macht:
Working fine here.
I was using your dev built on Odroid N2.
I installed old inputstreamhelper v0.5.2 to get new libwidevine.so 4.10.2252.0 installed.
Checked if libwidevine.so 4.10.2252.0 is used.
Played Netflix without problems.

Portisch:
By revert of https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/8f4205e7b2e8047b5907d75e003e7a459259672b we could confirm it is working again with 4.10.2252.0.

Thank you very much for the support!

Portisch:
@kszaq is it OK for you when I change the copyright to CoreELEC as LE is not go to merge it?
https://github.com/CoreELEC/CoreELEC/pull/287/commits/3ed379247e25acf81a9b9af5e86a5c53d00b494a#diff-80cbd4f0a3e38f8997c90f6ef8f4763732b2333d275b62300c7064870a4017a3R2

Or you feel more part of team LE?

kszaq:
Feel free to do whatever you need.

matthuisman:
@Portisch
Is it possible to back-port this to 9.2 branch?
Just so the users who want to stick with 18 or have old hardware (Odroid C2) can keep using Widevine content

@kszaq
I see LibreELEC preloads libtcmalloc.
But it seems CoreELEC doesn't and it still works fine.
I thought preload is required?
LibreELEC says you need to add the preload line to config for LE 10.
But if its not actually required - then why do the say it?
Unless I just can't spot where CoreELEC is preloading it..

oh, I see - there is some magic over on coreelec xbmc repo:
https://github.com/CoreELEC/xbmc/commit/4915c5f5c93d2e8f1e8b3507051799fd48da2e3a

Portisch:
It's planed in future, at least for **Amlogic** project. Maybe not for **Amlogic-ng** as here v19 can be used.
But we will wait first how it works for v19 to not break any stable project.

Portisch:
@kszaq @wagnerch: our users were just running into segmentation faults using the inbuilt tcmalloc in kodi.bin.

@vpeter4 found out that `LD_PRELOAD=/storage/.kodi/cdm/libwidevinecdm.so` in `kodi.conf` does work as well.
There is no need for tcmalloc anymore and no TLS block error.

Did you know or even tried this solution? Any experience?

kszaq:
@Portisch I've been running a custom build with tcmalloc preloaded (Kodi not linked to it) and I've had no issues - but my setup is quite minimal as I use Kodi exclusively for Widevine-protected content.

My **guess** is that if you link Kodi to tcmalloc, then use another library (binary addon) that is not linked to it, that another library uses "standard" malloc which conflicts with tcmalloc. If you preload tcmalloc, then all libraries loaded by Kodi use tcmalloc as it takes precedence over "standard" malloc.

But I may be wrong.

Preloading widevine was a first solution suggested in https://github.com/xbmc/inputstream.adaptive/issues/678#issuecomment-839287040

Using tcmalloc instead of preloading libwidevine was suggested as a "safer" approach, e.g. for people who have not yet downloaded widevine, it also allows using it without restarting Kodi etc. 

If you want to drop tcmalloc, I would suggest to add something like https://github.com/LibreELEC/LibreELEC.tv/pull/5399/commits/817f2bfa69571a21ba7acde35e679c498c8063ac and preload `/storage/.kodi/cdm/libwidevinecdm.so` if present.

kszaq:
Looking at symbols exported by libwidevine is should be safe to preload it:
```
# nm --defined-only -D libwidevinecdm.so
003fd79c T CreateCdmInstance
003fd788 T DeinitializeCdmModule
0080e680 T _fini
003fd934 T GetCdmVersion
0080e674 T _init
003fd760 T InitializeCdmModule_4
0080efc0 T __start_google_malloc
0080e688 T __start_malloc_hook
00812c70 T __stop_google_malloc
0080efbc T __stop_malloc_hook
003fd944 T VerifyCdmHost_0
```

Portisch:
None of the devs have seen any issue but the last release showed now some users reporting issues and rebooting devices.

I think another problem of preload libwidevinecdm.so is that inputstream.adaptive do open with dlopen but never close it again. The dlclose call is disabled in source. When now a stream get opened/closed multiple times the system freeze. I want to try first to use the library functions directly in inputstream.adaptive and remove the dlopen call. Maybe the use of tcmalloc can be avoided at all.

matthuisman:
will preload line if the file doesnt exist crash?
eg. if the user deletes the widevine lib?

**Also, I just tried - and it seems Google hasn't yet revoked the Chrome (ARM) versions.
4.10.1679.0 is still working for me on Pi4 LibreELEC 9.2.6**

Portisch:
No, LD_PRELOAD just ignore it if the file does not exist. It's save, but as kszaq said it's necessary to reboot to get cdm active.

matthuisman:
reboot will be ok. 
Inputstream Adaptive Helper could give a message saying reboot is required.

So, then all that required is the glib patch and the preload entry?
gperftools no longer needed?

And there is no way we can LD_PRELOAD from inside Inputstream Adaptive before we dlopen the Widevine blob?

kszaq:
@matthuisman The preload needs to be done for Kodi, so IA is unable to do this.

matthuisman:
did you ever try different dlopen flags once you had the glib patch in?
at moment, IA is using RTLD_LAZY. I wonder if RTLD_NOW or RTLD_GLOBAL or RTLD_DEEPBIND could remove the need to preload?
https://github.com/xbmc/inputstream.adaptive/blob/Matrix/wvdecrypter/cdm/base/native_library_posix.cc#L28

kszaq:
@matthuisman I only tried RTLD_DEEPBIND and it also resulted in TLS error. I did not try other options.

wagnerch:
LD_PRELOAD vs linking should have the same behavior.  libtcmalloc replaces malloc() in both cases.  I guess the advantage of LD_PRELOAD would be it's easily removed.  But I have been running on LibreELEC with libtcmalloc_minimal linked to kodi with no issues.  I have also been running it with CoreELEC, but where that device sits I do not use it as much.

I believe the TLS error is a result of the fact that tcmalloc replaces malloc.  There are a few notes/blurbs on tcmalloc files that mention LD_PRELOAD'ing tcmalloc can be problematic, which is why I personally went down the linking route.

http://goog-perftools.sourceforge.net/doc/tcmalloc.html
```To use TCmalloc, just link tcmalloc into your application via the "-ltcmalloc" linker flag.

You can use tcmalloc in applications you didn't compile yourself, by using LD_PRELOAD:

   $ LD_PRELOAD="/usr/lib/libtcmalloc.so" 

LD_PRELOAD is tricky, and we don't necessarily recommend this mode of usage.

TCMalloc includes a heap checker and heap profiler as well.

If you'd rather link in a version of TCMalloc that does not include the heap profiler and checker (perhaps to reduce binary size for a static binary), you can link in libtcmalloc_minimal instead.```

**Comment by matthuisman** [comment]
"you can link in libtcmalloc_minimal instead."

would be nice not having have to reboot after installing widevine.
Could CoreELEC just use what LibreELEC did? = preload libtcmalloc_minimal

**Comment by wagnerch** [comment]
RTLD_LAZY - means symbol resolution is deferred until the point of a dlsym call.
RTLD_NOW - means symbols are resolved at time of dlopen call.
RTLD_DEEPBIND - means symbol resolution will use this libraries (being dlopen'd) ahead of global symbols.

I don't think any of those options would have anything to do with TLS.  The TLS error is a memory allocation issue, and just doing  strings on libwidevinecdm it seems to me that libtcmalloc is statically linked into that shared library.

I think the other thing that did work, btw, was LD_PRELOAD on libwidevinecdm itself.  But that option requires restarting Kodi anytime the CDM changes, and not to mention if it does replace malloc with tcmalloc (because it's statically linked) then it wouldn't solve the crashing.  But there is one way to find out, try it.

In terms of the crashing, are there any backtraces from gdb?

**Comment by matthuisman** [comment]
Can't we just revert
https://github.com/CoreELEC/xbmc/commit/dcb336292adc6a2127dffcd283d027de1f867de7

then add the echo "LD_PRELOAD=/usr/lib/libtcmalloc_minimal.so" >> /run/libreelec/kodi.conf

just like LibreELEC?
it seems libtcmalloc_minimal  doesn't have the issues that libtcmalloc has (according to your quoted message above)

**Comment by vpeter4** [comment]
In CE Kodi is linked with same _libtcmalloc_minimal.so_ as would be used with preloading. There is no "full" libtcmalloc used.

**Comment by wagnerch** [comment]
@Portisch I am wondering if the issue is related to link order.  Apparently that can be an issue and was looking through the README for tcmalloc and it does suggest compiling (gcc) with the CFLAGS:
-fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free

https://chromium.googlesource.com/external/gperftools/+/master/README

Checking to see if that compiles.  LD_PRELOAD will also work, and it sounds like from other threads you had good results.  The plus is clearly that it's easy to remove.  It would seem to suggest, at least to me, that link order is a thing.

The flags mentioned above do compile, haven't tested it.  But LD_PRELOAD is probably the easiest & safest way to go.
https://github.com/wagnerch/LibreELEC.tv/commit/c3954431ab5218f83ab255b8b27bb08b83f8c9fa
https://github.com/wagnerch/xbmc/commit/1efeb33a759587d23a1622f45a92084a1c503122

**Comment by Portisch** [comment]
Thank you for this hint. It compiles with these flags. Maybe this is the difference between current linked in and LD_PRELOAD.

But first we need to define the reason what cause kodi crashes/rebooting as it also happen when using LD_PRELOAD.

**Comment by wagnerch** [comment]
@Portisch One recent post has this backtrace:

```
Thread 1 (Thread 0xe5a08240 (LWP 23571)):
#0  0xf73c3dac in tcmalloc::CentralFreeList::FetchFromOneSpans(int, void**, void**) () from /usr/lib/libtcmalloc_minimal.so
#1  0xf73c420c in tcmalloc::CentralFreeList::RemoveRange(void**, void**, int) () from /usr/lib/libtcmalloc_minimal.so
#2  0xf73c7d40 in tcmalloc::ThreadCache::FetchFromCentralCache(unsigned int, int, void* (*)(unsigned int)) () from /usr/lib/libtcmalloc_minimal.so
#3  0x0033d760 in NPT_HttpHeaders::AddHeader(char const*, char const*) ()
#4  0x0031396c in PLT_DeviceHost::SendSsdpSearchResponse(PLT_DeviceData*, NPT_HttpResponse&, NPT_UdpSocket&, char const*, NPT_SocketAddress const*) ()
#5  0x003271b8 in PLT_SsdpDeviceSearchResponseInterfaceIterator::operator()(NPT_NetworkInterface*&) const ()
#6  0x00327430 in PLT_SsdpDeviceSearchResponseTask::DoRun() ()
#7  0x00325a7c in PLT_ThreadTask::Run() ()
#8  0x00355ed0 in NPT_PosixThread::EntryPoint(void*) ()
#9  0xf737ad64 in start_thread (arg=0xe5a08240) at pthread_create.c:463
#10 0xf31911b8 in ?? () at ../sysdeps/unix/sysv/linux/arm/clone.S:73 from /usr/lib/libc.so.6
```

According to the net, this is a program bug caused by memory corruption like a double free.  Apparently glibc malloc()/free() may not detect, or has safeguards to ignore this type of program bug.  So perhaps it has to do with other addons users have or features they are using.  Or it is something specific to the Amlogic patches ported forward from Kodi 18.x, but seems good enough for me to watch 1 crappy movie and a 45 minute YouTube video using the 5/26 nightly.  :)  The only way to identify the cause of the bug is to actually remove tcmalloc, and link in [asan](https://github.com/google/sanitizers/wiki/AddressSanitizer) -- which will use glibc malloc.  (From what I can tell, asan might be llvm specific, so maybe valgrind or something like that will help find any memory allocation issues).

No issue for me on LibreELEC w/ RPi4 either.  My setup is pretty basic, Tvheadend for PVR, Netflix, Youtube, and my personal movie/tvshows collections.  No special skins, shared MySQL db.

So in short, yes the crashing does seem to be caused by tcmalloc, but tcmalloc may be detecting an actual bug.

**Comment by vpeter4** [comment]
> According to the net, this is a program bug caused by memory corruption like a double free. Apparently glibc malloc()/free() may not detect, or has safeguards to ignore this type of program bug.

Well, this doesn't look ok to me? Glibc detects double free but when using libtcmalloc_minimal.so PRELOAD-ed not?

```
vim3:~ # cat c_test.c
int main() {
  void *p = malloc(1024);
  for (int i=1; i<=1000; i++) {
      printf("free #%d\n", i);
      free(p);
  }
}

vim3:~ # ./c_test
free #1
free #2
double free or corruption (!prev)
Aborted

vim3:~ # LD_PRELOAD=/usr/lib/libtcmalloc_minimal.so ./c_test
free #1
free #2
free #3
........
free #998
free #999
free #1000
```

Update1: Ok, this happens with no extra cflags. Let's use the same as Kodi is using.
When this program is build with `-O3` (or with any optimization flag) then glibc doesn't detect double free anymore.

**Comment by kszaq** [comment]
@wagnerch This looks like a bug related to UPnP and I can confirm that Kodi crashed when I started browsing the library over UPnP...

Meanwhile I've been running Kodi without tcmalloc and with Widevine preloaded - no crashes so far.

**Comment by wagnerch** [comment]
@kszaq Good to know.  Maybe some folks having trouble can turn that off and see if it is stable.  I was looking in that code last night, but didn't see anything obvious.  As for the reason for tcmalloc, it was mainly to work around the TLS error which can also be worked around by preloading widevinecdm -- but that is somewhat undesirable and requires a restart when the CDM is updated.  But considering we haven't seen a true widevine update in like 2 years, maybe it's not a big deal.

The reason I assume tcmalloc is the culprint is that it's known to be problematic when dlopen'd, it's known to cause this TLS error when dlopen'd, it replaces malloc which is something that occurs when a library is dlopen'd and TLS is allocated, and it was evident from strings that tcmalloc is statically linked in libwidevinecdm.  No direct evidence of anything, just an assumption.  If we can figure out another solution for this TLS error then that would be the way to go, I guess.

@vpeter4 Good to know the behavior is similar for a double free.  I threw out double free as a potential example, that was one of the examples found as the cause for the segmentation fault in that function, but it was generally stated as memory/heap corruption.  I believe another example was freeing unreferenced memory, for example an uninitialized pointer (not malloc'd) but free'd anyways, depending on optimization level the pointer may be referencing unknown memory.

**Comment by Portisch** [comment]
We removed tcmalloc and LD_PRELOAD of the CDM library itself for now. We get segment faults on any attempt. So currently no new CDM lib support.

**Comment by wagnerch** [comment]
It looks like the upnp library used in Kodi is no longer maintained, and did see a pull request on their github that actually does affect this specific area where one user reported a backtrace.  Basically the issue was a buffer overflow due to an over-length HTTP header.  Apparently underlying NPT_Strings is the class used to allocate headers, and it is backed by NPT_Size (uhh which should be enough, 32-bit unsigned int) but there is definitely re-allocation happening in that class when an existing header is modified.

https://github.com/axiomatic-systems/Neptune/pull/10
```
the following changes fix a buffer overflow in String.append and set an upper limit to the size of a single HTTP-Header in order to improve robustness.

The buffer overflow results if there is an overflow in NPT_Size (32bit int) as the Reserve function does not update if the new size is smaller. So data is written in unused mem
```

vpeter4:
@wagnerch: Behaviour is not similar - tcmalloc minimal doesn't detect double free regardless of the optimization mode.

matthuisman:
@Portisch 
Maybe time to get the backport to Leia going?
Google have started reverting the widevine last night.

vascobraga41:
> @Portisch
> Maybe time to get the backport to Leia going?
> Google have started reverting the widevine last night.

Hi Matt, 
We're now already focused on Nexus and will look into backporting things to Leia after this initial work is done. We encourage those who can to update to Matrix, that's pretty stable!

PS - where have you detected the reversion, 4.10.1679 still works here with Netflix and Disney plus.

base08:
4.10.1679 Not working as well here… Damn you google…

matthuisman:
@vascobraga41 
HBO Europe failing. Basically all NZ addons (most use Brightcove). Googles own widevine server also not working.
Foxtel Go Australia down too. And probably a lot more but don't yet have time to install them all on my test pi to check

But yes, Disney and Netflix still seem ok for now

vascobraga41:
So I tested the only 2 working, that's lucky...

Edit - confirmed HBO go isn't working.

base08:
It’s a shame as I’m stuck in Leia (s912 device) and can’t use my widevine dependent addons anymore…

Portisch:
> It’s a shame

What is a shame? Pushing distribution to support old hardware/software by doing this for free in their free time?
Please feel free and provide a fully qualified PR to backport the cdm support for project Amlogic with 3.14 kernel.

Project Amlogic (GXM & GXBB) is closed and obsolete. Every change might break the current perfect stability and so any change must be tested carefully and need also to be qualified. And all this done for a obsolete system...

Buying 4K TVs and Dolby Atmos AVR and every year a new iPhone, but do not spend a few bucks on a cheap media player equipped with like a S905X3. It's a shame, yes!

matthuisman:
@Portisch 
i don't think he is asking for Kodi 19 for S912 (big job)
Think they just would like a new 9.2.X (kodi 18) build that supports latest Widevine.

I've just did a PR that brings in what I think all the required changes that LibreELEC 9.2 did
https://github.com/CoreELEC/CoreELEC/pull/295

Hopefully it's as simple as that.
Just to get Kodi 18 S912 users "back in the game"
I'd also much appreciate it, as they have no upgrade path to 10.0 (due to the big kernel changes etc needed)

matthuisman:
@base08 
I built an update .tar for S905/S912 with the the PR changes above.

You can get it below:
https://sourceforge.net/projects/coreelec/files/CoreELEC-Amlogic.arm-9.2-devel-1630714868.tar/download
That will get Widevine working until (hopefully) CoreELEC release 9.2.8

base08:
@matthuisman thank you for your initiative! Much appreciated and like me thousand others will hopefully benefit from your quick patch. Will test and post here the results…

PR #288, opened 2021-05-24
Amlogic-ce: add ondemand tuning script from LE/Amlogic

kszaq:
The script adjusts ondemand tunables to make CPU scale more aggresively, resulting in smoother GUI and possibly some minor improvements when SW-decoding. Copied from upstream projects/Amlogic.


kszaq:
NB: this won't work until kernel config is switched to boot with ondemand.

Portisch:
It's possible for the user to change the governor behavior by himself:
https://github.com/CoreELEC/service.coreelec.settings/blob/coreelec-19/src/resources/lib/modules/hardware.py#L313

Default is performance as ondemand was leading to issues at users side.

PR #289, opened 2021-06-05
gobject-introspecion: fix building for host

kszaq:
Without this patch, the build fails with
```
configure: error: C compiler cannot create executables
```
the cause is
```
/usr/bin/ld: error while loading shared libraries: libbfd-2.35.1.so: cannot open shared object file: No such file or directory
```

libbfd.so installs to `$TOOLCHAIN/$HOST_NAME/$TARGET_NAME`

### DIFF
--- a/packages/devel/gobject-introspection/package.mk
+++ b/packages/devel/gobject-introspection/package.mk
@@ -34,7 +34,6 @@ EOF
 }
 
 pre_configure_host() {
-  export LD_LIBRARY_PATH+=":$TOOLCHAIN/lib"
   CFLAGS+=" -fPIC"
 }
 


vpeter4:
Can you rebuild this package without change above, tar whole folder `build.CoreELEC-Amlogic-ng.arm-19/build/gobject-introspection-1.54.1` and upload it somewhere?
Because this package builds for all of us and it doesn't link with bfd library anyway.

Also on what host you are building?

kszaq:
Host is a VM running Ubuntu 20.04.2 LTS minimal, used only to build CE19 and LE10. Host provides binutils 2.34.

The same behaviour was observed on Ubuntu 16 minimal, used for the same purposes.

Without this patch gobject-instrospection does not build, here is the build folder:
[gobject-introspection-1.54.1.tar.zip](https://github.com/CoreELEC/CoreELEC/files/6604861/gobject-introspection-1.54.1.tar.zip)

From config.log:
```
configure:3770: /home/kszaq/CoreELEC/build.CoreELEC-Amlogic-ng.arm-19/toolchain/bin/host-gcc -march=native -O2 -Wall -pipe -I/home/kszaq/CoreELEC/build.CoreELEC-Amlogic-ng.arm-19/toolchain/include -Wno-format-security -fPIC  -Wl,-rpath,/home/kszaq/CoreELEC/build.CoreELEC-Amlogic-ng.arm-19/toolchain/lib -L/home/kszaq/CoreELEC/build.CoreELEC-Amlogic-ng.arm-19/toolchain/lib conftest.c  >&5
/usr/bin/ld: error while loading shared libraries: libbfd-2.35.1.so: cannot open shared object file: No such file or directory
```

vpeter4:
And LE10 fails too? Then send PR to LE because CE merges things from there.

I'm making one VM to try this out. I don't see real reason why bfd lib is needed in this case.

vpeter4:
I tried building on fresh 20.04.2 LTS and have no issues. Not sure why in your case host gcc needs libbfd for a simple program like hello world.

Also others also use Ubuntu 20 without issues.

kszaq:
On both my 16.04 and 20.04 VMs I used a netboot ISO + installed packages suggested by `checkdeps` to have bare minimal to compile.

vpeter4:
I made another VM with mini.iso and repeat the steps: everything ok. I would conclude something is wrong with your build environment.

arthur-liberman:
If you are using Windows 10, I strongly recommend setting up WSL2.
It works great, and builds CE without any issues. Only thing to note, is you probably want to limit the maximum amount of RAM available to WSL2, as the build process can eat it all up and slow your system down to a crawl.

kszaq:
1. Install Ubuntu Mini, all options at default, from http://archive.ubuntu.com/ubuntu/dists/focal/main/installer-amd64/current/legacy-images/netboot/mini.iso
2. `sudo apt install git build-essential`
3. `git clone --depth 1 https://github.com/CoreELEC/CoreELEC.git`
4. `cd CoreELEC ; ./scripts/checkdeps`
5. `./scripts/build gobject-introspection`

Result: build failure.

What am I doing wrong?

vpeter4:
Well, usually we do

```
cd CoreELEC
make
```

Seems gobject-introspection host is missing some dependency if build directly. Will do some more investigation.

kszaq:
@vpeter4 Issue is the same if `make` is used to build a whole image.

kszaq:
Removing `export LD_LIBRARY_PATH` from `package.mk` also fixes building for me.

vpeter4:
Finally I had a success to replicate this issue :-)
Not sure why previously just build it. But now I follow exactly your step above.
Also not sure why it works for everybody.

From what i see /usr/bin/ld linker from the host links with toolchain/lib/libctf.so.0. And this lib needs libbfd-2.35.1.so. Because it doesn't find it it fails.

So correct fix would be to remove LD_LIBRARY_PATH. Please squash both commits and I think it will be merged.

Portisch:
@kszaq please squash the two commits together that we can merge it, thank you.

kszaq:
Squashed and rebased.

PR #290, opened 2021-06-30
Widevine: replace glibc version suffix solution to ENV variable

dtechsrv:
Replace the original glibc version suffix (*-arm64tls) to an environment variable.
This is latest solution in script.module.inputstreamhelper addon to detect that the glibc package already contains the TLS 64-byte alignment patch.

References:
https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/450
https://github.com/emilsvennesson/script.module.inputstreamhelper/pull/457


Portisch:
Thank you for your support.

Please include the remove of this line as well:
https://github.com/CoreELEC/CoreELEC/blob/coreelec-19/packages/devel/glibc/package.mk#L16

As soon a new inputstreamhelper release is available we can merge it, thank you.

PR #294, opened 2021-08-25
Fix luna build issue, update packages

TheChoconut:
This should fix building issue where moonlight-embedded would detect X11 as a dependency even though it shouldn't.

It also bumps luna to the latest version and removes the old addon that is no longer maintained (to remove any confusions).

If there is any way to compile the addons without merging to main branch, I would recommend testing it, though on my build VM nothing was broken.


Portisch:
Should this moonlight addon removed at LE as well?
So is it needed to upstream the commit?

TheChoconut:
I'll be using script.luna for Kodi addon. I recommend removing it if luna is working for Team LE, since it's only a dead addon saying "Moonlight is no longer maintained", which I'm currently trying to maintain and this IMO makes it confusing.

CvH:
`Should this moonlight addon removed at LE as well`
its a empty addon that creates kind of a placeholder at the addon.xml that everyone who has this addon installed get a notification that the addon was removed, see https://github.com/LibreELEC/LibreELEC.tv/blob/master/config/addon/xbmc.broken.xml.

Its the kodi way of removing broken addons.

If you plan on reviving the moonlight addon its like easier to just edit the moonlight/package.mk.
So if someone has an old moonlight installed get updated with the new one too.

PR #295, opened 2021-09-03
Support latest widevine with glibc patches (Portisch, wagnerch, dtechsrv)

matthuisman:
just copied in these changed from Libreelec 9.2 / CoreELEC 20 done by the below authors.
@Portisch, @kszaq , @dtechsrv 

hopefully someone can check / build / test and confirm all is ok.

There was 1 commit that I didn't bring across from LibreELEC:
https://github.com/LibreELEC/LibreELEC.tv/commit/e4d6559bfdb3b92a636cac30f27781d1496970c1
I couldn't see that code in CoreELEC 20 so assume not needed for CoreELEC

### DIFF
--- a/packages/mediacenter/kodi/scripts/kodi-config
+++ b/packages/mediacenter/kodi/scripts/kodi-config
@@ -66,6 +66,11 @@ else #arm
   echo "MALLOC_MMAP_THRESHOLD_=8192" >> /run/libreelec/kodi.conf
 fi
 
+# required for inputstreamhelper to detect TLS 64-bytes support
+if [ -n "$(uname -m | grep arm)" ] || [ "$(uname -m)" == "aarch64" ]; then
+  echo "LIBC_WIDEVINE_PATCHLEVEL=1" >> /run/libreelec/kodi.conf
+fi
+
 if [ -f /storage/.config/kodi.conf ] ; then
   cat /storage/.config/kodi.conf >>/run/libreelec/kodi.conf
 fi
--- a/packages/devel/glibc/patches/aarch64/glibc-add-support-for-SHT_RELR-sections.patch
+++ b/packages/devel/glibc/patches/aarch64/glibc-add-support-for-SHT_RELR-sections.patch
@@ -1,0 +1,348 @@
+From 6676e967dba405ca31d57b63e096becd13d4a200 Mon Sep 17 00:00:00 2001
+From: Rahul Chaudhry <rahulchaudhry@chromium.org>
+Date: Thu, 15 Mar 2018 14:30:17 -0700
+Subject: [PATCH 4/8] sys-libs/glibc: add support for SHT_RELR sections.
+
+This patch adds experimental support for SHT_RELR sections, proposed
+here: https://groups.google.com/forum/#!topic/generic-abi/bX460iggiKg
+
+SHT_RELR sections are supported for arm, aarch64, and x86_64 targets.
+To enable them, pass '--experimental-use-relr' flag to gold.
+
+Definitions for the new ELF section type and dynamic array tags, as well
+as the encoding used in the new section are all under discussion and are
+subject to change. We plan to send the patch upstream after the gABI has
+been updated to include the new definitions.
+
+[Adrian: forward-ported to glibc 2.32]
+---
+ elf/do-rel.h                 | 41 ++++++++++++++++++++++++++++++++++--
+ elf/dynamic-link.h           | 15 +++++++++++++
+ elf/elf.h                    | 15 +++++++++++--
+ elf/get-dynamic-info.h       |  7 ++++++
+ sysdeps/aarch64/dl-machine.h | 10 +++++++++
+ sysdeps/arm/dl-machine.h     | 10 +++++++++
+ sysdeps/i386/dl-machine.h    | 10 +++++++++
+ sysdeps/x86_64/dl-machine.h  | 10 +++++++++
+ 8 files changed, 114 insertions(+), 4 deletions(-)
+
+diff --git a/elf/do-rel.h b/elf/do-rel.h
+index 1d0a1f2c5d..25babef6e1 100644
+--- a/elf/do-rel.h
++++ b/elf/do-rel.h
+@@ -26,6 +26,12 @@
+ # define elf_machine_rel_relative	elf_machine_rela_relative
+ #endif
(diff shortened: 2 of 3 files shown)


Portisch:
@matthuisman thank you for your support

The current patch sub directory `arm` is ignored in 9.2 branch. This got fixed later in CE-19 and above but not in CE-9.2.
The patches should apply only for project **Amlogic**, not for **Amlogic-ng** as latest can be upgraded to CE-19 or CE-20. GXBB and GXM users instead have no possibility to upgrade to Amlogic-ng.

So please change the glibc patches location to the sub directory like `amlogic-3.14` and add `PKG_PATCH_DIRS` in glibc package.mk like this example:
```c
diff --git a/packages/devel/glibc/package.mk b/packages/devel/glibc/package.mk
index cf5acd603c..065a55d30b 100644
--- a/packages/devel/glibc/package.mk
+++ b/packages/devel/glibc/package.mk
@@ -45,6 +45,10 @@ else
   PKG_CONFIGURE_OPTS_TARGET="$PKG_CONFIGURE_OPTS_TARGET --disable-debug"
 fi
 
+if [ "${PROJECT}" = "Amlogic" ]; then
+  PKG_PATCH_DIRS+="amlogic-3.14"
+fi
+
 pre_build_target() {
   cd $PKG_BUILD
     aclocal --force --verbose
```

packages/mediacenter/kodi/scripts/kodi-config
```diff
@@ -66,6 +66,11 @@ else #arm
   echo "MALLOC_MMAP_THRESHOLD_=8192" >> /run/libreelec/kodi.conf
 fi
 
+# required for inputstreamhelper to detect TLS 64-bytes support
+if [ -n "$(uname -m | grep arm)" ] || [ "$(uname -m)" == "aarch64" ]; then
```
Portisch, line 70: `LIBC_WIDEVINE_PATCHLEVEL` should also only apply to project Amlogic. This can be done like with:
```suggestion
if [ "$(uname -r)" = "3.14.29" ]; then
```
CE do have only `arm` so there is no need to check it and project **Amlogic** can be detected by the kernel version.

dtechsrv: I thought all ARMs were affected, including ng-builds. Why should it be limited to the amlogic-3.14 kernel only?

dtechsrv:
> The current patch sub directory `arm` is ignored in 9.2 branch. This got fixed later in CE-19 and above but not in CE-9.2.

Patches will never be loaded from the "arm" directory if the kernel is "aarch64".

Here is the solution for this problem:
https://github.com/LibreELEC/LibreELEC.tv/pull/5455/commits/e4d6559bfdb3b92a636cac30f27781d1496970c1

packages/mediacenter/kodi/scripts/kodi-config
Portisch:
```
...Amlogic-ng as latest can be upgraded to CE-19 or CE-20...
```
...which include all needed fixes for already 3 month now. Leia support is stopped since a while from TK. Matrix is done and stable and Nexus shows up in next time.

Portisch:
Yes, this was a problem and got fixed. So for 9.2 it needs to be 'aarch64' (default) or any other folder defined by `PKG_PATCH_DIRS`. But as the patch should be applied only for project Amlogic it should not be placed in `aarch64` as it would get applied for all projects.

packages/mediacenter/kodi/scripts/kodi-config
dtechsrv: Okay, but it still affects all ARMs, including 64-bit kernels (like: aarch64).

This will not make it mandatory to build official images for the Amlogic-ng devices, but anyone who wants to do this for themselves should have their source patched in the same way. 
For example: LE didn't build their 9.2.8 images other than PRi2/3/4. It is the same situation.

dtechsrv:
> But as the patch should be applied only for project Amlogic it should not be placed in `aarch64` as it would get applied for all projects.

Well, my patch does the exact opposite, so you don't need a separate directory for `aarch64`.
```
# workaround to use arm patches for aarch64
if [ "${TARGET_PATCH_ARCH}" = "aarch64" ]; then
  PKG_PATCH_DIRS="arm"
fi
```
Nothing needs to be placed in `aarch64` because that directory doesn't even exist, it uses the `arm` patch directory only.
Under my own project (LibreELEC-AML) the Meson8* (amlogic-3.10: 32-bit arm kernel) and the GXBB/GXL (amlogic-3.14: 64-bit aarch64 kernel) work simultaneously without any further modifications outside of this patch on the branch of LE-9.2.

matthuisman:
ok, i moved the patches into a folder called aarch64 so they are actually detected in 9.2.

It seems a bit odd to actively try to disable / bypass stuff for -ng users just to force upgrade to 19?
But i'll do it if we must to get a new build out.

If they are happy with kodi 18, why not at least allow them to build themselves and for it to work?
Or some 3rd party could host it somewhere. Even I would if needed.

Or coreelec could release 9.2.8 for all the same platforms as 9.2.7 and get some big kudos!

The changes are minimal after-all. It won't be like your supporting a whole new release.
You know they are stable as the 19 -ng has had them in for some time now.

It'll probabably cost less time than explaining to every -ng user when widevine stuff stops playing that they must go to 19.
Then do the back and forth about some add-ons breaking etc.
Saves us all time if they have an easy path to stay on a working 18 if they really want to.

I've done a pr here: https://github.com/CoreELEC/CoreELEC/pull/296
which has your changes you suggested.
Please just merge whichever you like, and ill close the other

Portisch:
Thank you for your changes.

We need to discuss this internal as such changes do include some other items like support. Like what happens if in some months something similar does change and break the system? How long this can be done, maintain multiple branches and releases?

The main reason of this discussion is the step to python 3  done in Matrix and that some users do not want to upgrade as their addons aren't working or maintained anymore. But this step needs to be done some day anyway. So why not now?

Please be patient, thank you.

otava5:
@matthuisman 

Please how can I apply the changes to an already installed CE?

matthuisman:
@otava5 
https://discourse.coreelec.org/t/new-widevine-solution-backport-to-leia/16667/8

Portisch:
This closes #296

PR #298, opened 2021-10-10
Update RTL8822BU wifi driver

ishioni:
The upstream repo for this driver has changed. But moreover, this driver in combination with a TP-Link T3U stick produces kernel oops and disconnections. I've compiled and tested CE-19.2 on a Odroid C4 with this attached version, and the crashes seem to be gone


ishioni:
If this gets merged, should I also make a separate PR for coreelec-20 branch?

Portisch:
No, not needed, thx. I will pick it for branch coreelec-20 when it's merged.
Thank you for your support!

PR #302, opened 2021-11-11
Enabled sqlite ICU support.

Kurganov:
Enabled sqlite ICU support. This make case insensitive like statement for non ASCII characters.
+ Fix nfs-utils depends.


Portisch:
What is the reason the `icu` package got moved?
`sqlite` build fine in same place.

Kurganov:
Yes, build fine, but more logical place for the package is in `packages/textproc` than in `packages/addons/addon-depends/`. Or is it preferable to be backward compatible with libreelec?

Portisch:
Maybe it's better to upstream it to LE/master at all?
If it get merged it will be automatic included in CE as well.

Portisch:
Any news on this PR @Kurganov ?

Kurganov:
a lot of work before holidays, not time to test LE build. After 13 dec.

Kurganov:
add PR to LE/master

PR #309, opened 2022-03-14
fixed oe_setup_addon unsafe with user inputs from addons settings

spleen1981:
If user input in an addon setting text field includes a double quote, call to `oe_setup_addon` will fail with error, as current function tries to `eval $(var="someinputwith"doublequotes")` which includes additional unescaped double quote, resulting in unfinished string.
Proposed modification uses `sed` to single quote all strings and escape possible single quotes included in original input, avoiding also possible security issues.


Portisch:
Thank you for your support!

It's the wrong file, correct one would be:
`projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf`

spleen1981:
> Thank you for your support!
> 
> It's the wrong file, correct one would be: `projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf`

modification moved to the correct file

PR #310, opened 2022-03-14
[CoreELEC19] fixed oe_setup_addon unsafe with user inputs from addons settings

spleen1981:
If user input in an addon setting text field includes a double quote, call to `oe_setup_addon` will fail with error, as current function tries to `eval $(var="someinputwith"doublequotes")` which includes additional unescaped double quote, resulting in unfinished string.
Proposed modification uses `sed` to single quote all strings and escape possible single quotes included in original input, avoiding also possible security issues.

### DIFF
--- a/projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf
+++ b/projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf
@@ -31,9 +31,9 @@ oe_setup_addon() {
       if [ -f "$xml_file" ]; then
         XML_SETTINGS_VER="$(xmlstarlet sel -t -m settings -v @version $xml_file)"
         if [ "$XML_SETTINGS_VER" = "2" ]; then
-          eval $(xmlstarlet sel -t -m settings/setting -v @id -o "=\"" -v . -o "\"" -n "$xml_file")
+          eval $(xmlstarlet sel -t -m settings/setting -v @id -o "=" -v . -n "$xml_file" | sed -e "s/'/'\\\\''/g; s/=/='/; s/$/'/")
         else
-          eval $(xmlstarlet sel -t -m settings -m setting -v @id -o "=\"" -v @value -o "\"" -n "$xml_file")
+          eval $(xmlstarlet sel -t -m settings -m setting -v @id -o "=" -v @value -n "$xml_file" | sed -e "s/'/'\\\\''/g; s/=/='/; s/$/'/")
         fi
       fi
     done


Portisch:
Thank you for your support!

It's the wrong file, correct one would be:
`projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf`

spleen1981:
> Thank you for your support!
> 
> It's the wrong file, correct one would be: `projects/Amlogic-ce/packages/mediacenter/kodi/profile.d/00-addons.conf`

modification moved to the correct file

vpeter4:
I think this solves the problem same way without using sed (which is not needed in my opinion):
`eval $(xmlstarlet sel -t -m settings/setting -v @id -o "='" -v . -o "'" -n "$xml_file")`

Original command prints **id**, then **="**, then **value**, then **"** character and at the end **new line**. Your proposal is using single quotes instead of double quotes. That's why just use single quote with xmlstarlet durectly.

Ah, now I see: problem is still single quote in value :)

spleen1981:
> I think this solves the problem same way without using sed (which is not needed in my opinion): `eval $(xmlstarlet sel -t -m settings/setting -v @id -o "='" -v . -o "'" -n "$xml_file")`
> 
> Original command prints **id**, then **="**, then **value**, then **"** character and at the end **new line**. Your proposal is using single quotes instead of double quotes. That's why just use single quote with xmlstarlet durectly.

sed is needed to escape possible single quotes in user input.
Hence no enclosing single quotes are added directly with xmlstarlet itself, all possible occurencies are escaped with sed and enclosing  single quotes are added with sed at the end.
If you use xmlstarlet to enclose with single quotes only, without the sed part, and an odd number of single quotes are included in user input, you will get `syntax error: unterminated quoted string`

vpeter4:
Yes, I though this could be fixed in xmlstarlet but seems not. I don't like first sed expression but we have to live with it.

vpeter4:
Just one more question: Why do you replace single quote `'` with `'\''` ? Why not just `\'` ?

spleen1981:
> Just one more question: Why do you replace single quote `'` with `'\''` ? Why not just `\'` ?

You can't escape single quote that way, you need to close the previous single quote, escape this single quote, and open a new one.
if you want to achieve `single'quote` string you need to `'single'\''quote'`: if you `'single\'quote'` you will have a single quoted `single\` string, a not quoted `quote` string  and the last single quote alone, with consequent `syntax error: unterminated quoted string`

vpeter4:
Complicated.

Anyway, I'm good with this changes.

vpeter4:
Complicated.

Maybe it is just me but wouldn't this looks better? Looking slash and backslash is confusing...
`sed -e "s|'|'\\\\''|g; s|=|='|; s|$|'|"`

Anyway, I'm good with this changes.

spleen1981:
> Complicated.
> 
> Maybe it is just me but wouldn't this looks better? Looking slash and backslash is confusing... `sed -e "s|'|'\\\\''|g; s|=|='|; s|$|'|"`
> 
> Anyway, I'm good with this changes.

Custom delimiters are needed in case of conflicts with patterns rather than for cosmetic use... personally I like to stick with default notation if possible.
However it doesn't look like more clear or readable with pipes instead of slashes to me, but I guess it's subjective...

PR #311, opened 2022-03-30
Coreelec 19 | Shanwan Gamepad Hid module

hbiyik:
Hello

This PR adds a gamepad support for the below mentioned device. It can not be an external package to the old 4.19 kernel, and needs to modify usb-core, hid-core modules both, since 4.19 does not support plug and play hid modules.

Deatils are below.
https://github.com/hbiyik/hid-shanwan

This patch also works fine with corelec-20, if its ok i can PR to corelec-20 branch.


Portisch:
Please make your commits directly to linux repo. There is no need to use the outdated method by carry patch files.
https://github.com/CoreELEC/linux-amlogic

Please follow the naming rules for the commits so they can be identified more easy like: `HID: sony:...`

When your patch is merged does it interfere somehow with this [commit](https://github.com/CoreELEC/linux-amlogic/commit/a874da03bd5768056602b0eb1717ef0de56ca2ea)?
Or can it be dropped at all?

hbiyik:
The mentioned commit is for ps3 controller clones, where they directly clone the sony controllers. This is for pc based contorollers which is totally different 

Amlogic 4.19 kernel is totally different than upstream kernel input drivers. Therefore upstreaming the patch would not help to amlogic kernel, details in the link.

Anyways, i understand if you dont want to maintain those things, i can always patch my own CE. But this controllers are really high runners in aliexpress.

Portisch:
No, it's ok. We will not upstream it to Amlogic as we have no possibility at all ;)
https://github.com/CoreELEC/linux-amlogic is our repo based on Amlogic 4.9 kernel.

So it would be no problem for us to merge a PR at the linux repo.
We try to handle as less patch files as possible as it results only in more maintenance with the need of someone doing this job.
It's much easier if the commits are directly merged into the repo.

hbiyik:
Ah ok got it, ill PR to the linux repo then, thanks for the guide.

PR #313, opened 2022-06-24
nfs-utils: use dynamic mirror url

7Ji:
Sourceforge uses various mirrors to distribute packages according to users' GeolP. The mirror list is [here](https://sourceforge.net/p/forge/documentation/Mirrors/) and it could be updated. Sometimes some mirror server can even be removed completely and render the package not downloadable at all if we use a URL with that mirror, like the old iweb mirror that was used broken a couple days ago until fixed in https://github.com/CoreELEC/CoreELEC/commit/a473c38648ffdd5c1cde6835c89f929715d43b0d

Different mirros have different accessibility depending on user's actual location and some could even be inaccessible for users outside of their serving areas (e.g. A user [reports on EE forum](https://emuelec.discourse.group/t/controller-profile-issue-with-standalone-cores-4-5/1177/23?u=7ji) that the netix mirror currently being used is inaccessible for him). The current netix mirror in the package.mk is located in Bulgaria and has bad performance for some of my test servers (China, US, JP. But RU is OK.)

The ``SOURCEFORGE_SRC`` variable is defined in [config/source](https://github.com/CoreELEC/CoreELEC/blob/coreelec-20/config/sources), and is used in other packages that also downloads from sourceforge like [pcre](https://github.com/CoreELEC/CoreELEC/blob/coreelec-20/packages/devel/pcre/package.mk). 

Using this as the constructor of URL instead, the URL created for the current version will be https://prdownloads.sourceforge.net/nfs/nfs-utils/2.6.1/nfs-utils-2.6.1.tar.bz2 instead. And it will 302 redirect to the correct mirror according to IP by sourceforge, that'll be both available and fastest. So we won't need to worry that the netix mirror we current use could go down like iweb one day.


vpeter4:
Please change also for coreelec-19 branch.

PR #315, opened 2022-07-06
(emegrcy!)lose support of rpc.h/ubuntu22.04

u0-ani-nya:
due to libc v2.32 removed rpc.h and instead of provide tirpc lib, we must use command line --enable-tirpc.
but it didn't work at main makefile but only works when configure at nfs-utils folder.
should we detect libc version and auto fill the command line?


vpeter4:
Wrong repo and also bad PR ?

u0-ani-nya:
> Wrong repo and also bad PR ?

dev repo
image

u0-ani-nya:
i tried manually configure with tirpc enabled， it works.
but our program seems don't wanna skip this bug. it just re-configure it
so the bug is unskipable -- again and again

vpeter4:
This is CoreELEC not EmuELEC.
And just use Ubuntu 20 like we all do and you will not have such issues.

u0-ani-nya:
i will try. thank you
might some times feauture, it hopes will be fixed.
i don't know why the libc team do so.

vpeter4:
Maybe this would work
```
diff --git a/packages/network/nfs-utils/package.mk b/packages/network/nfs-utils/package.mk
index 5b785d8ee6..7cf97a522c 100644
--- a/packages/network/nfs-utils/package.mk
+++ b/packages/network/nfs-utils/package.mk
@@ -43,6 +43,9 @@ pre_configure_host() {
     --disable-tirpc \
     --without-systemd \
     --without-tcp-wrappers"
+
+  # for Ubuntu 22 with libntirpc-dev package installed
+  CFLAGS+=" -I/usr/include/tirpc"
 }

 pre_configure_target() {
```

u0-ani-nya:
thank you a lot ！！！

PR #319, opened 2022-08-18
fix the dep error for fedora >= 34

biaji:
After fedora 34, the bdftopcf has been move it's own rpm named "bdftopcf".  Which cause a broken of dep check.


vpeter4:
Send this to https://github.com/LibreELEC/LibreELEC.tv

biaji:
> 

Right away! Thanks.

PR #323, opened 2022-12-10
Update luna and moonlight dependencies

TheChoconut:
It's been a while since I've last had a good update... Now that some upstream changes have fixed issues, seems like a good time to update luna.


Portisch:
Is the PR finished and do it compile?

TheChoconut:
It compiles for me using `create_addon` (although it tells me an error about circular dependency on openssl, I think that's upstream?)

Portisch:
Is this also required for CE-19? Then please add a new PR for it, thank you very much.

PR #324, opened 2022-12-10
support for the Sega Model 2 and Model 3

sleepingkarma:
Will it be possible to have support for the Sega Model 2 and Model 3?


Portisch:
Wrong repo, and a PR is not a issue!

PR #326, opened 2022-12-20
initramfs: host depends on fakeroot

kszaq:
When building single package from a clean checkout, build fails because of non-present `fakeroot`. Fix this by adding `fakeroot:host` to depends.


Portisch:
Maybe you want to upstream it here?https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/virtual/initramfs/package.mk
When it's in LE `master` branch it will be merged into `coreelec-20` later on.

kszaq:
This patch is CE-specific, as LE doesn't use `fakeroot` in initramfs package. 🙂

PR #327, opened 2022-12-30
Amlogic-ne: enable DVB drivers

smp79:


Portisch:
Please change commit name to like `Amlogic-ne: enable...`
Otherwise it will be confusing when the branch get merged.

PR #329, opened 2023-01-20
update 8822BU to v5.13.1

ishioni:


Portisch:
I am not sure if it's ok to switch completely to another repo as it looks completely different.
Can not be the patch applied to the original used source?
Or is this feature not already be implemented?

ishioni:
That's the problem with RTL drivers, it's up to random community members to try and upkeep them. To be honest, I'm having problems compiling CE-20 right now, with weird docker errors and broken urls for packages (addon.joystick being one). Where can i find the stuff you guys use for compiling the official release? 

As for features - this supposedly makes WPA3 work for one, and should be a bit more stable - the old version of the driver could do kpanics

vpeter4:
joystick package was just updated.

ishioni:
I will try building again in the evening, will come back with results

vpeter4:
We are using driver RTL8821CU from morrownr too.

ishioni:
Ok, i tried again, the problem with addon.joystick is indeed fixed, but I still can't build. At the very last step, when packing the rootfs, it errors out, because it's trying to set permissions, but can't because the files are root:root. I think it should use sudo but doesn't?

sudo is of course present and usable

```
Creating 4.0 filesystem on /build/target/CoreELEC-Amlogic-ng.arm-20.1-Nexus_devel_20230120215611.system, block size 524288.
[..]
Running /build/projects/Amlogic-ce/devices/Amlogic-ng/bootloader/release                                                              
cp: preserving permissions for 'target/CoreELEC-Amlogic-ng.arm-20.1-Nexus_devel_20230120215611/3rdparty/bootloader/Generic_cfgload': Operation not permitted
[..]
*********** FAILED COMMAND ***********                                                                                                
. ${FOUND_PATH}                                                    
**************************************                                                                                                
make: *** [Makefile:10: image] Error 1
```

vpeter4:
> I think it should use sudo but doesn't

Whole build process runs without sudo requirement. Didn't saw error like this so I can't help much. Just run `make` and create .tar file only.

ishioni:
> > I think it should use sudo but doesn't
> 
> Whole build process runs without sudo requirement. Didn't saw error like this so I can't help much. Just run `make` and create .tar file only.

Just tried it like that, from scratch, same result. Weird, i built CE9 in exactly the same way before, and never had this problem :/
Screenshot 2023-01-21 at 11 30 10

ishioni:
@Portisch @vpeter4 I have revisted this and rebased it on top of current coreelec-20 branch. It compiled sucesfully and is running wonderfull on my Odroid C4 with a TP-Link adapter. Could you please merge?
Screenshot 2023-08-17 at 13 39 49
Screenshot 2023-08-17 at 13 38 28

PR #331, opened 2023-03-23
RTL8821CU: bump package to 20210708

zcmi:
update


Portisch:
Please use more details in the commit message like: `RTL8821CU: bump package to 20210708`

PR #333, opened 2023-04-01
add support to build on manjaro as well as arch and endeavouros

pellcorp:
I would like to be able to build coreelec and downstream repos on manjaro, since endeavouros is already supported as a arch derivative, I am hopeful this simple MR will be accepted as well.


Portisch:
Better upstream the PR?
https://github.com/LibreELEC/LibreELEC.tv/blob/master/scripts/checkdeps

PR #335, opened 2023-10-25
use 32bit entware installer on amlogic-ng-dv

geropan:
amlogic-ng-dv build uses 64bit entware installer, which fails to install via the installentware script.

This PR changes entware installer relevant architecture.


vpeter4:
Doesn't amlogic-ng also has wrong arch set?

arthur-liberman:
The normal NG build uses aarch64 kernel, so I guess that it can run either version.

PR #338, opened 2023-11-17
homatics-leds: add german translation

ghost:
@vpeter4 
I have made some effort and added a german language.


vpeter4:
Thank you.

PR #341, opened 2023-11-27
Amlogic-ce: delete project-specific tvheadend43

kszaq:
This is an RFC pull request. 

Project-specific TVHeadend 4.3 addon can be removed because upstream LE is now more up-to-date. I haven't noticed any Amlogic-specific patches so it should be fine to use the updated rev. 

With this PR, the addon compiles fine for `-ng` and works with USB Xbox DVB-T/C tuner.


kszaq:
Currently, there are 3 versions of `tvheadend43` in CE Addons, but addon version is lower with every newer build:
21.0.105 older than 21.0.1, older than 21.0.0:
https://github.com/CoreELEC/CoreELEC-Addons/tree/Amlogic-ng/21.0/arm/service.tvheadend43

This is due to ADDON_VERSION bumped upstream with PKG_REV reset. CE has not bumped ADDON_VERSION, so many addons have more recent builds with lower version number:
https://github.com/CoreELEC/CoreELEC-Addons/tree/Amlogic-ng/21.0/arm/service.mariadb
https://github.com/CoreELEC/CoreELEC-Addons/tree/Amlogic-ng/21.0/arm/service.minidlna
https://github.com/CoreELEC/CoreELEC-Addons/tree/Amlogic-ng/21.0/arm/service.system.docker

and so on.

PR #344, opened 2023-12-29
Replace Amlogic VNC server package with an improved version

dtechsrv:
I modified the original work of @kszaq at several points and added many new things, such as improved image buffer comparison, mouse support, screen resolution change detection, etc.

The full changelog of the version included in the package is available here:
https://github.com/dtechsrv/aml-vnc-server/releases/tag/1.1.0

This package is prepared for the CoreELEC build system following its previous model, but many things had to be changed due to the modification of the aml-vnc binary.

_I only tested under `Amlogic-ng` because I don't have a device compatible with `Amlogic-ne`._

### DIFF
--- a/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
+++ b/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
@@ -1,3 +1,9 @@
+21.0.5
+- replace package source git repository
+- creating a source directory with the necessary files and their modifications
+- modify the service start script to be compatible with the new binary
+- add Hungarian translation of the add-on
+
 21.0.4
 - performance fixes
 
--- a/projects/Amlogic-ce/packages/addons/service/aml-vnc/package.mk
+++ b/projects/Amlogic-ce/packages/addons/service/aml-vnc/package.mk
@@ -2,13 +2,13 @@
 # Copyright (C) 2018-present Team CoreELEC (https://coreelec.org)
 
 PKG_NAME="aml-vnc"
-PKG_VERSION="c34903e68695110c6af5df6fb43797950a0f335b"
-PKG_SHA256="650b664c7fb402f37bb9f2001f3a26bc85ad04d9eb0f2c273c446514cc3c12f4"
-PKG_REV="4"
+PKG_VERSION="1.1.0"
+PKG_SHA256="50a0040b46019c2781f671401907be4b24fb9f5749a7e439c4111c0853a4fe58"
+PKG_REV="5"
 PKG_ARCH="arm aarch64"
 PKG_LICENSE="GPL"
-PKG_SITE="https://github.com/CoreELEC/aml-vnc"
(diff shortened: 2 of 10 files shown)


vpeter4:
Thanks for the PR. It will be merged after I test it.

vpeter4:
Please also make strncpy changes to prevent buffer overflows.

https://github.com/CoreELEC/aml-vnc/commit/e18f13e70cffb16ca29c421eac67435cc13e16a1

dtechsrv:
Oops, your indication about the buffer overflow is valid, but I prefer to use `snprintf()` instead, it also solves the null terminator problem by itself.

```
diff --git a/vncserver.c b/vncserver.c
index 719c6d4..c72202e 100644
--- a/vncserver.c
+++ b/vncserver.c
@@ -184,9 +184,9 @@ int main(int argc, char **argv) {

        // Preset values from environment variables (However, the values specified in the arguments have priority.)
        if (getenv("VNC_SERVERNAME"))
-               strcpy(VNC_SERVERNAME, getenv("VNC_SERVERNAME"));
+               snprintf(VNC_SERVERNAME, sizeof(VNC_SERVERNAME), "%s", getenv("VNC_SERVERNAME"));
        if (getenv("VNC_PASSWORD"))
-               strcpy(VNC_PASSWORD, getenv("VNC_PASSWORD"));
+               snprintf(VNC_PASSWORD, sizeof(VNC_PASSWORD), "%s", getenv("VNC_PASSWORD"));
        if (getenv("VNC_PORT"))
                VNC_PORT = atoi(getenv("VNC_PORT"));

@@ -206,11 +206,11 @@ int main(int argc, char **argv) {
                                                break;
                                        case 'n':
                                                i++;
-                                               strcpy(VNC_SERVERNAME,argv[i]);
+                                               snprintf(VNC_SERVERNAME, sizeof(VNC_SERVERNAME), "%s", argv[i]);
                                                break;
                                        case 'p':
                                                i++;
-                                               strcpy(VNC_PASSWORD,argv[i]);
+                                               snprintf(VNC_PASSWORD, sizeof(VNC_PASSWORD), "%s", argv[i]);
                                                break;
                                        case 'f':
                                                i++;

```

^ If you see this as good, I will send the correction soon.

vpeter4:
Sure, whatever you prefers.

dtechsrv:
The fix commit is submitted, the tag is corrected, and the hash is replaced in the package. So I'm done.

vpeter4:
All good and ready for merging.

But I do want to hear one more opinion from you. What I don't like with addons like this is that when some settings are changed for the addon the systemd service must be manually restarted. Little annoying thing .... So I propose something like this in `default.py` file which restarts the service on settings changed.

dtechsrv:
This is a very good idea, but unfortunately I'm not a Python expert at all, so I'm not sure I should implement this.

vpeter4:
Ok, then something for later :)

vpeter4:
I noticed that there are other addons with similar file and I copied one of them for this addon.

PR #347, opened 2024-02-12
aml-vnc: bump package to 1.2.0

dtechsrv:
List of changes since v1.1.0:
https://github.com/dtechsrv/aml-vnc-server/releases/tag/1.2.0


vpeter4:
Please set PKG_REV to 0.

dtechsrv:
But the earlier ones already have a greater number under this branch.
If I reset it now, the last package (21.0.5) will not be updated, because 0 is less.

Or am I thinking wrong?

Portisch:
I removed the commit as it was my fault. This `PKG_REV` handling is a mess and in your case it need to be bumped to `6`, yes.
There are other addons require to need to be reset to `0` when version is bumped.

PR #348, opened 2024-04-02
Enable USB printer support

kszaq:
Build USB printer module to allow using CE box as a CUPS or p910nd print server.


kszaq:
@Portisch Can you possibly have a look? 🙂

Portisch:
Sure, we are currently just really busy...

kszaq:
Thank you!

PR #349, opened 2024-04-16
Don't disable ipv6 support for nfs mounts

iteratee:
The default image supports ipv6, so there isn't a good reason to hobble nfs mounts any more, if there ever was.

### DIFF
--- a/packages/network/nfs-utils/package.mk
+++ b/packages/network/nfs-utils/package.mk
@@ -66,7 +66,6 @@ pre_configure_target() {
     --enable-tirpc \
     --enable-uuid \
     --disable-gss \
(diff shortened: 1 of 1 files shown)


vpeter4:
Thank you. I hope this was tested before sending PR.

iteratee:
> Thank you. I hope this was tested before sending PR.

Yes. I wanted to mount my v6-only NFS server and eventually figured out what the problem was. It's currently running w/o issue in my network.

The same change should likely be made everywhere applicable, like rsync.

vpeter4:
Ok, thank you.

If you are going to fix some other packages first check if also applies to LibreELEC and fix it there.

PR #350, opened 2024-04-16
Add support for rtl8723ds wifi adapter

mich2200:
Amlogic-ng: add new Wifi driver RTL8723DS


Portisch:
And please choose a better commit title like: `Amlogic-ne: add new Wifi driver RTL8723DS`
And in what CoreELEC project it should be included? Amlogic-ng, Amlogic-ne, all?

mich2200:
thank you for your reply and i think now it's cleared enough

vpeter4:
Thank you. I fixed the package myself.

PR #354, opened 2024-12-04
shell.service: close splash-image before starting shell

kszaq:
CoreELEC's `splash-image` is closed by `kodi.bin` which is not started in `textmode`. Use `=-` to prevent failure in case splash screen is disabled by `nosplash`

### DIFF
--- a/packages/sysutils/busybox/system.d/shell.service
+++ b/packages/sysutils/busybox/system.d/shell.service
@@ -5,6 +5,7 @@ After=multi-user.target
 [Service]
 Environment=TTY=1
 WorkingDirectory=/storage
+ExecStartPre=-/bin/killall -s SIGUSR1 splash-image
 ExecStartPre=/bin/sh -c 'echo -en "\033[?25h"'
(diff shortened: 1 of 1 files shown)


Portisch:
We do not see any usage of a media hardware without media center started.
But this service will only be called if starting in `textmode`. So save to merge.

kszaq:
My 2 use cases:
* Box connected to TV: start in `textmode`, use a script to monitor CEC bus and start Kodi only when TV is on - stop Kodi when turning TV off. That way the box can be always-on with low power usage and act as a TVHeadend server and print server
* Headless, low-cost and low-power (0.8W) server for some local services, CoreELEC and LibreELEC are a great superlightweight base for that kind of systems. I could have used any other distro available but most of the times they are either too heavy to my taste or don't offer read-only system with portable overlay such as CE/LE

Portisch:
And using `nosplash` does not work?

kszaq:
It does but if someone wants pretty splash and then console on the screen... 😄

PR #355, opened 2024-12-26
aml-vnc: update libvncserver to 0.9.15 and addon (22.0.0)

dtechsrv:
This PR contains the update of libvncserver to the current version, adds the missing alias for the aml-vnc service, and hides the password in the addon's configuration interface.

_Unfortunately, I cannot personally test the CE-22 version because I do not have a compatible device._


dtechsrv:
_I backported it to CE-21 and noticed a strange thing: the aliases in the systemd scripts don't work for some reason, but it doesn't seem to work for other services either._ 

_For example, 'systemctl stop docker' does not work for Docker. It doesn't show any error, but it has no effect. However, 'systemctl stop service.system.docker' works flawlessly._

_The strange part is that if I start it with 'systemctl start docker', then 'systemctl stop docker' works, but if I try to stop it with the full service name then it doesn't work._

**EDIT:** This systemd bug seems to be present in LE12 as well.
https://github.com/LibreELEC/LibreELEC.tv/issues/9618

vpeter4:
Thank you: checked on CE-22 NO Ugoos AM8.

PR #356, opened 2024-12-28
[CE-21] aml-vnc: update libvncserver to 0.9.15 and addon (21.1.1.7)

dtechsrv:
This is a backport of #355.

Tested with new CE-21 build on an X96 box (S905X/2GB).
```
CoreELEC (community): 21.1.1-Omega_devel_20241228123716 (Amlogic-ng.arm)
       Machine model: Amlogic
      CoreELEC dt-id: gxl_p212_2g
       Amlogic dt-id: gxl_p212_2g
       Linux version: 4.9.269 (dtech@BuildServer-11) #1 Sat Dec 28 14:10:11 CET 2024
       Kodi compiled: 2024-12-28 14:36:12 +0000
```


vpeter4:
Please remove the systemd fix because it will come in anyway from LE.

dtechsrv:
Won't the numbering cause problems? There, the two patch numbers are `0400` and `0500`, but because you already have an [0500](https://github.com/CoreELEC/CoreELEC/blob/coreelec-19/packages/sysutils/systemd/patches/systemd-0500-Add-working-directory-storage-for-debug-shell.patch), I had to rename them to `0600` and `0700`.

If it does come in from there, you will have two different patches with the same `0500` number, which, although it usually does not cause problems (especially if there is no overlap between them), but it can change the order of the patches.

If this does not cause a problem for you, I will of course remove it, but I would definitely like a confirmation.

~~Also, in this case, due to the operation of the service alias, you should wait for the systemd patches with the merge of this PR.~~
**EDIT:** _Because users will probably get the addon update sooner than CE-21.2 might be released, so it doesn't matter.
Although in the latter case it would be better to change the version of the addon from 21.1.1.7 to 21.2.0 anyway._

vpeter4:
The order of the patches doesn't matter in this case because our 500 is independent from the other 2 for alias fix.

And I doubt this alias fix really matter for users. In my opinion it is just a little annoying thing which doesn't cover all the time spend to fix it.

dtechsrv:
According to your request, I removed the systemd patches from the PR, so it is practically equivalent to #355 (except for the addon versioning).

PR #364, opened 2025-10-03
fix pcsc protocol in oscam addon

mczerski:
pcsc plugin for communication with pcscd service in oscam service was failing due to missing shared library. This PR brings changes from corelec-22 branch into coreelec-21


Portisch:
oscam is maintained by upstream: https://github.com/LibreELEC/LibreELEC.tv/blob/master/packages/addons/service/oscam/package.mk

mczerski:
FYI https://github.com/LibreELEC/LibreELEC.tv/pull/10552

PR #376, opened 2026-03-12
aml-vnc: bump package to 1.4.0_beta8

dtechsrv:
There has been some new options added into last betas:
- You can now disable sending mouse actions if you don't need it.
- The libvncserver debug log is also disabled by default, making the journal more readable.
- Added a soft reinitialization option to prevent connection loss when screen size does not change.

_The addon icon has also been replaced, because it was previously used UltraVNC's icon._


dtechsrv:
The new logo design was this:
aml-vnc_new2_le

But, if the red logo suits CE better, I can replace it with this:
aml-vnc_new2_ce

vpeter4:
Blue is fine I think.

PR #379, opened 2026-05-06
aml-vnc: bump package to 1.4.0_beta11

dtechsrv:
Headless mode assumtion changed to optional in this beta, because a working solution should be found for the new (S905X5+) and older (S905X4 and earlier) SoCs as well. As I know in the case of the new SoCs, FBDEV is not actually used by Kodi.

Moreover, since headless mode does not work on newer SoCs, the mouse used over the bootsplash is actually controlling Kodi, but completely blindly, so this mode should be prevented at the system level.

### DIFF
--- a/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
+++ b/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
@@ -1,5 +1,5 @@
 22.0.0
-- bump package to 1.4.0_beta10
+- bump package to 1.4.0_beta11
 - change addon icon
 - add new config options
 
--- a/projects/Amlogic-ce/packages/addons/service/aml-vnc/package.mk
+++ b/projects/Amlogic-ce/packages/addons/service/aml-vnc/package.mk
@@ -2,9 +2,9 @@
 # Copyright (C) 2018-present Team CoreELEC (https://coreelec.org)
 
 PKG_NAME="aml-vnc"
-PKG_VERSION="1.4.0_beta10"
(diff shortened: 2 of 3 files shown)


vpeter4:
> this mode should be prevented at the system level

What do you mean by this? What needs to be done? Checked the SoC?

Portisch:
As the tearing of GBM GUI is now fixed it's not said that SoCs before S7D will stay with FBDEV.
They might will also switch to libMali GBM when FBDEV get no updates anymore.

The only way to be sure is to load libMali.so and check if `gbm_create_device` is exported.
This is only the case if it's a GBM libMali.

dtechsrv:
> What do you mean by this? What needs to be done? Checked the SoC?

All we need to do is create an `if` statement that includes the indicated line if it is true. That's all I wanted to ask.

> As the tearing of GBM GUI is now fixed it's not said that SoCs before S7D will stay with FBDEV. They might will also switch to libMali GBM when FBDEV get no updates anymore.

Then there is no point in further supporting "headless fallback" to FBDEV on the VNC server side, because then it will not work at all. The currently used DRM framebuffer capture method doesn't work without a display, unfortunately that's a simple fact. The most obvious solution is to use a dummy HDMI stick (instead of a physical display), but there's no real point in this software workaround on the server side.

At least that's what I think. What do you think about this?

Portisch:
Smart! Does it work with auto detection now?

dtechsrv:
> Smart! Does it work with auto detection now?

Yes, it works perfectly and only allows FBDEV fallback when supported.

I started with a simple test originally, which looked like this:
ce22-backend-test

If you also think this is appropriate, then you can merge. Later, if FBDEV is completely dropped for older SoCs, this condition check can be deleted from the service startup script.

PR #381, opened 2026-05-13
Amlogic: reset Dolby Vision start-decode state on seek

Tristan1019-user:
## Summary

This resets the Amlogic Dolby Vision bitstream converter's start-decode state after seek/reset discontinuities for Dolby Vision and FEL/MEL streams.

## Why

Amlogic Dolby Vision playback can stall after seeking when the converter keeps stale start-decode state across a discontinuity. The existing reset path already handles AV1 and VP9. This extends the same kind of start-decode reset to Dolby Vision/FEL/MEL while preserving the converter metadata and cached RPU/SPS/PPS state.

## Validation

Checks performed:

- Patch dry-run passed against `kodi-98ba3144f46b8e1bda4c547520b8eab998593071` source.
- `git diff --check` passed.
- Downstream test build was used for Dolby Vision/FEL playback validation on Fire TV Cube 2nd Gen.
- Long DV/FEL playback was stable in the downstream build, and seek-focused testing was used while investigating this path.

This PR intentionally contains only the codec reset behavior. It does not include AVDV process-info UI work, HDR conversion controls, or device-specific FireCube patches.

### DIFF
--- a/projects/Amlogic-ce/packages/mediacenter/kodi/patches/kodi-999.80-amlogic-dovi-reset-start-decode-on-seek.patch
+++ b/projects/Amlogic-ce/packages/mediacenter/kodi/patches/kodi-999.80-amlogic-dovi-reset-start-decode-on-seek.patch
@@ -1,0 +1,31 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Tristan Stiller <stillertristan77@gmail.com>
+Date: Tue, 12 May 2026 00:00:00 -0500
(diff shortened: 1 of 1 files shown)


Portisch:
Wrong repo, use https://github.com/CoreELEC/xbmc

Tristan1019-user:
You are right, this belongs in CoreELEC/xbmc rather than the CoreELEC package repo. I moved the same source-side change there as CoreELEC/xbmc#46: https://github.com/CoreELEC/xbmc/pull/46

PR #382, opened 2026-05-13
runtime: create var paths before setup writes

Tristan1019-user:
## Summary

This trims the runtime-directory hardening to the two setup paths that write directly into `/var` runtime locations:

- Ensure `/var/run` exists before `tz-data.service` links `/var/run/localtime`.
- Ensure `/var/lib` exists before `libmali-overlay-setup` creates the active `/var/lib/libMali.so` symlink.

## Scope notes

Earlier changes for ConnMan, OpenSSH, and `/run/libreelec` were removed because those paths are already covered by existing tmpfiles entries or service setup.

This PR is intentionally narrow and does not change boot ordering or add broad runtime-directory policy. It only creates the target directory immediately before each service/script writes into it.

## Validation

Checks performed:

- Verified the PR diff now contains only `tz-data.service` and `libmali-overlay-setup`.
- `git diff --check origin/coreelec-22` passed.
- Downstream CE22 integration build using these changes booted with `kodi=active`, `connman=active`, and working Wi-Fi.

### DIFF
--- a/packages/sysutils/tz/system.d/tz-data.service
+++ b/packages/sysutils/tz/system.d/tz-data.service
@@ -8,6 +8,7 @@ After=var.mount systemd-tmpfiles-setup.service
 Type=oneshot
 Environment=TIMEZONE=UTC
 EnvironmentFile=-/storage/.cache/timezone
+ExecStartPre=/bin/mkdir -p /var/run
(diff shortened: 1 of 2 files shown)


vpeter4:
No need AI generated solutions without actual problem and tested before pushing PR.

PR #383, opened 2026-05-13
nfs-utils: fix nfsrahead udev helper path

Tristan1019-user:
## Summary

This fixes the installed `nfsrahead` udev helper path in the CoreELEC `nfs-utils` packaging.

`nfs-utils` 2.9.x installs `nfsrahead` under the target libdir, while the installed udev rule can still point at `/usr/libexec/nfsrahead`. This rewrites the rule to call `/usr/lib/nfsrahead`, matching where the helper is installed in the target image.

## Scope notes

The earlier version of this PR also added manual `rpc-statd` runtime path creation. That was removed because `packages/network/nfs-utils/tmpfiles.d/nfs-utils.conf` already creates:

- `/run/nfs/sm`
- `/run/nfs/sm.bak`
- `/run/nfs/rmtab`
- `/run/nfs/etab`

So this PR is now intentionally limited to the `nfsrahead` path mismatch only.

## Validation

Checks performed:

- Confirmed the existing tmpfiles config already owns the `/run/nfs` state paths.
- Confirmed the PR no longer modifies `rpc-statd.service`.
- Confirmed the remaining diff only updates the installed `99-nfs.rules` helper path from `/usr/libexec/nfsrahead` to `/usr/lib/nfsrahead`.
- `git diff --check` passed.

### DIFF
--- a/packages/network/nfs-utils/package.mk
+++ b/packages/network/nfs-utils/package.mk
@@ -111,6 +111,12 @@ post_makeinstall_target() {
   chmod 755 ${INSTALL}/sbin/*
   mv ${INSTALL}/sbin/* ${INSTALL}/usr/sbin
   rmdir ${INSTALL}/sbin
+
+  # nfs-utils 2.9.x installs nfsrahead under libexecdir, but its udev rule
+  # still references /usr/libexec on our target. Keep the helper and rule in
+  # agreement so bdi events do not fill boot logs with failed PROGRAM calls.
+  sed -i 's|/usr/libexec/nfsrahead|/usr/lib/nfsrahead|g' \
+    "${INSTALL}/usr/lib/udev/rules.d/99-nfs.rules"
 }
 
 post_install() {


vpeter4:
What is the point of creating folders and files manually if there is `tmpfiles.d/nfs-utils.conf` ?

Tristan1019-user:
You're right; I missed that `tmpfiles.d/nfs-utils.conf` already creates those `/run/nfs` directories/files. I removed the `rpc-statd.service` `ExecStartPre` changes and force-pushed the branch. The PR is now limited to the remaining `nfsrahead` udev helper path mismatch (`/usr/libexec/nfsrahead` -> `/usr/lib/nfsrahead`). Thanks for catching that.

vpeter4:
Come on, why was then even added? And what exactly this PR fixes?

Tristan1019-user:
Fair question. The `rpc-statd` part was added while debugging a downstream boot/runtime issue and I did not notice that CoreELEC already covers those paths via `packages/network/nfs-utils/tmpfiles.d/nfs-utils.conf`. That part was wrong for this PR and is now removed.

What remains is only the `nfsrahead` udev helper path mismatch:

- `nfs-utils-2.9.1/tools/nfsrahead/99-nfs.rules` ships `PROGRAM="/usr/libexec/nfsrahead %k"`.
- CoreELEC configures the target package with `--libexecdir=/usr/lib`.
- The helper is installed as `/usr/lib/nfsrahead`, not `/usr/libexec/nfsrahead`.

So without this change, the installed udev rule can point at a helper path that does not exist. The PR now only rewrites the installed rule to match CoreELEC's configured libexec path. If you prefer this as a source patch against `tools/nfsrahead/99-nfs.rules` instead of a post-install `sed`, I can change it that way.

vpeter4:
Are you messing with me? Unsquashed build 20260513:
```
$ cat squashfs-root/etc/os-release 
NAME="CoreELEC"
VERSION="22.0-Piers_nightly_20260513"

$ ll squashfs-root/usr/lib/nfsrahead
-rwxr-xr-x 1 pvic pvic 68632 May 13 00:46 ./squashfs-root/usr/lib/nfsrahead*

$ cat squashfs-root/usr/lib/udev/rules.d/99-nfs.rules
SUBSYSTEM=="bdi", ACTION=="add", PROGRAM="/usr/lib/nfsrahead %k", ATTR{read_ahead_kb}="%c"
```

https://github.com/CoreELEC/CoreELEC/blob/5e7303a7d35979c94062df733766910c4c034897/packages/network/nfs-utils/package.mk#L69

https://git.linux-nfs.org/?p=steved/nfs-utils.git;a=blob;f=tools/nfsrahead/99-nfs.rules.in;h=648813c556bac8132815592944058d2367e5ac58;hb=HEAD

PR #384, opened 2026-05-13
Amlogic: wait for HDMI DRM connector before first window init

Tristan1019-user:
## Summary

This adds a bounded Amlogic DRM settle/reprobe path before Kodi initializes the first window against HDMI.

Changes included:

- Delay first window initialization briefly when HDMI modes are visible but the DRM connector still reports disconnected.
- Avoid creating the GUI against `dummy_l` or a half-bound scanout when the HDMI encoder/CRTC has not settled yet.
- Clean DRM object pointers on reprobe and initialize display state after the connector is actually connected.
- Treat hotplug events with a disconnected connector as not-ready instead of forcing a bad mode selection.

## Why

On a CE22 / kernel 5.15 Amlogic-no test build, Kodi could start while HDMI modes existed but the DRM connector was still reported disconnected. In the bad path Kodi fell through into an invalid first GUI/display state. Waiting briefly for the connector/encoder to bind avoids that first-start race without inventing fake connector or CRTC state.

## Scope notes

This is not FireCube-specific in the code or comments. The issue was observed on a Fire TV Cube 2nd Gen downstream build, but the fix is framed as a generic Amlogic DRM initialization race because the failure mode is in the Amlogic HDMI DRM/Kodi window initialization path.

## Validation

Checks performed:

- Both Kodi patch files dry-run cleanly against `kodi-98ba3144f46b8e1bda4c547520b8eab998593071` source.
- `git diff --check` passed.
- The downstream integration build using this behavior booted to Kodi without the previous first-start display failure.
- The deployed test build reported `kodi=active`, `connman=active`, and working network after boot.

This PR is separated from FireCube device-tree work and AVDV UI work so the DRM behavior can be reviewed independently.

### DIFF
--- a/projects/Amlogic-ce/packages/mediacenter/kodi/patches/kodi-999.85-amlogic-initial-drm-settle.patch
+++ b/projects/Amlogic-ce/packages/mediacenter/kodi/patches/kodi-999.85-amlogic-initial-drm-settle.patch
@@ -1,0 +1,54 @@
+--- a/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
++++ b/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
+@@ -169,5 +169,15 @@ void CWinSystemAmlogic::HotplugEvent()
+ {
+   m_amlDisplay->aml_init_drmDevice();
++  drmModeConnection connection;
++  int mode_count = m_amlDisplay->aml_get_display_modes_count(&connection);
++  if (connection == DRM_MODE_DISCONNECTED)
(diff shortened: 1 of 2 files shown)


Portisch:
Wrong repo, please use https://github.com/CoreELEC/xbmc

Tristan1019-user:
You are right, this is Kodi/windowing source code and belongs in CoreELEC/xbmc. I moved it there as CoreELEC/xbmc#47: https://github.com/CoreELEC/xbmc/pull/47

PR #385, opened 2026-05-16
aml-vnc: bump package to 1.4.0_beta13

dtechsrv:
**Changes:**
https://github.com/dtechsrv/aml-vnc-server/compare/1.4.0_beta11...1.4.0_beta13

Due to switch to dual framebuffer in the S7D DRM implementation (and probably also under S5 and S6), the detection of the current buffer was modified on server side as well, because in the case of a resolution or frequency change, it often did not detect and use the right source buffer that was really used for the capture, which produced blank screen. _(I don't know when the DRM stack was changed, but the framebuffer pairs didn't exist two months ago, so it gave me a bit of a headache over this weekend.)_

In addition, I replaced the assumption of headless mode with forced use of FBDEV for older SoCs, which is usually much better solution here than using the DRM backend directly.

### DIFF
--- a/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
+++ b/projects/Amlogic-ce/packages/addons/service/aml-vnc/changelog.txt
@@ -1,5 +1,5 @@
 22.0.0
-- bump package to 1.4.0_beta11
+- bump package to 1.4.0_beta13
 - change addon icon
 - add new config options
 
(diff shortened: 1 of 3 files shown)


dtechsrv:
Since the implementation of the reardonia patches, the DRM stack has been a mess in Kodi. What was stable in VNC server last week is now "partially" broken again.

Does it make sense for me to tinker with buffer management at all right now, or should I wait a few weeks/months?

Portisch:
I am not sure, I tried to use this "improvement" for CoreELEC. But it does break GUI colours and transparency. This may be caused because we use already hardware to transform the gui on the fly to like B2020. So it would be applied twice.

This new HDR FBO will not be used in near future I guess as it's just not need fo Amlogic.
The current PQ method is working well.

I forgot to merge the bump on last rebase, will include it on next one!

dtechsrv:
In the meantime, I modified the buffer handling, and beta13 now handles buffer change events much more flexibly, so what was a rare issue in beta12 has now been completely fixed.

PR #386, opened 2026-05-23
kodi: add patch to fix subtitle extradata discarded in CDVDDemuxClient

GenkaiToppa:
Adds the one-line fix from [xbmc#28362](https://github.com/xbmc/xbmc/pull/28362), currently open against Kodi master.

`CDVDDemuxClient::SetStreamProps` silently drops every subtitle stream's `extradata` unless it is exactly 4 bytes. ASS/SSA script headers are several kilobytes long, so they are always dropped, and `CDVDOverlayCodecSSA::Open` fails with `Unable to init overlay codec` for every InputStream addon that delivers rich subtitles. The patch replaces the size-equality gate with the same non-empty check the audio and video branches already use. The 4-byte DVB-subtitle identifier still passes, so PVR is unaffected.

Reproduced and fixed in a local Kodi 21.x build (same code path, master has it at line 528 instead of 522). Stock: error and no subtitles render. Patched: `CDVDSubtitlesLibass: Creating new ASS track`, ASS renders correctly.

Can be dropped once xbmc#28362 lands and kodi is bumped past it in CoreELEC.


Portisch:
Please user https://github.com/CoreELEC/xbmc for PR as we avoid to handle patches.

PR #387, opened 2026-05-23
kodi: add backport patch to fix subtitle extradata discarded in CDVDDemuxClient

GenkaiToppa:
Adds the one-line fix from [xbmc#28362](https://github.com/xbmc/xbmc/pull/28362) (currently open against Kodi master; an Omega cherry-pick PR will follow).

`CDVDDemuxClient::SetStreamProps` silently drops every subtitle stream's `extradata` unless it is exactly 4 bytes. ASS/SSA script headers are several kilobytes long, so they are always dropped, and `CDVDOverlayCodecSSA::Open` fails with `Unable to init overlay codec` for every InputStream addon that delivers rich subtitles. The same code is present in 21.x at line 522. The patch replaces the size-equality gate with the same non-empty check the audio and video branches already use. The 4-byte DVB-subtitle identifier still passes, so PVR is unaffected.

Reproduced and fixed in a local Kodi 21.2-Omega build. Stock: error and no subtitles render. Patched: `CDVDSubtitlesLibass: Creating new ASS track`, ASS renders correctly.

Can be dropped once the Omega cherry-pick has been merged and kodi is bumped past it in CoreELEC.


Portisch:
Closed because of https://wiki.coreelec.org/coreelec:eolomega

PR #399, opened 2026-06-20
Add back key mappings for various remote controls

jamal2362:
This fixes the issue where the OK button on the video OSD does not open on some third-party remote controls.


vpeter4:
You do know it was removed here? https://github.com/CoreELEC/CoreELEC/commit/d42f361e18f7524902b9f7e456be7c59eb25c9f5

```
Those keys are now properly handled by kodi and remapping them to KEY_ENTER / KEY_C is no longer necessary
```
Looks like this part is not working correctly?

jamal2362:
Yes, I'm familiar with that commit.

I'm using a G20S Pro BT remote, connected with the included RF USB dongle. General navigation and the OK button work as expected, but the OK/Select action does not work in the Video OSD. When I'm in the video OSD and press the OK button, nothing happens. 

I understand that this mapping was removed in LibreELEC. After adding the hwdb file back into the hwdb.d directory, the behaviour is restored and the Video OSD can be controlled with the OK button again, just like before.

So from my perspective, the removed key mapping appears to be the reason for the issue.

vpeter4:
> the removed key mapping appears to be the reason for the issue

True, but seems solution is somewhere else (?in Kodi)

Portisch:
it was this commit https://github.com/CoreELEC/xbmc/commit/68fa0eac44b42cf2cf4fd0417a67d3b8afdd5607
by 
https://github.com/CoreELEC/CoreELEC/commit/d42f361e18f7524902b9f7e456be7c59eb25c9f5
Done by "very tested" systems it seems...

CE-Repo:
@vpeter your fix build works fine. 
OK Button works again in video osd. Thank you!

https://discourse.coreelec.org/t/nightly-builds-new/8409/2750

Can I close here, then?

Portisch:
Closed by https://github.com/CoreELEC/xbmc/commit/15841057a0fca57fd961ee538bea3a41d1a2944a

CE-Repo:
Great thank you!

---

I know this is not directly related to the current topic, but I don't know where else to report it. Is there an email address or another preferred way to report bugs?

The media type labels used to be displayed correctly: SDR content was shown as SDR, HDR content as HDR, and Dolby Vision content as DV. However, for the last few nightly builds, everything is being reported as SDR, even though my library contains HDR10 and Dolby Vision media.

Unfortunately, I can't pinpoint exactly when this started happening. Under Windows, the same media files are detected and displayed correctly, so the issue appears to be specific to CoreELEC.

---

Also, starting today, there's been a micro stuttering issue with the AM9 Pro when I open TinyPPI and watching an DV file. That wasn't the case before. It looks like this was triggered by the GBM change. It was still working fine with yesterday's nightly build. 

Could someone take a look at this?

Thanks!

Portisch:
Use CoreELEC forum.

PR #405, opened 2026-07-12
CE-21: aml-vnc: bump package to 1.4.0

dtechsrv:
This is a backport to the current stable version.

**Changes:**
https://github.com/dtechsrv/aml-vnc-server/compare/1.3.0...1.4.0

Since the `libdrm` dependency is not included in the package, the binary is compiled with FBDEV support only.


Portisch:
CE-21 have no GBM support, so only FBDEV.

dtechsrv:
> CE-21 have no GBM support, so only FBDEV.

I know, I just wanted to indicate the difference between the two `package.mk`.

PR #408, opened 2026-07-29
aml-dtbtools: bump package to b2ca13ce

wyatt-paxiv:
Bumps the CE-local aml-dtbtools build tool (dtbTool/dtbSplit) from `cce100f` to `b2ca13ce`.

Two upstream commits:
- `b2ca13ce` Add missing header for Linux compilation
- `557fde5e` Minor formatting fix

Also switches `PKG_VERSION` to the full 40-char revision per the project's STANDARDS.md (`cce100f` was a grandfathered short hash). SHA256 verified against the downloaded tarball.


vpeter4:
Don't just bump packages without actually building CE. I was sceptical about `SHA256 verified against the downloaded tarball` because when package is downloaded it automatically show the SHA. No need to manually download and compare SHA.

And in this case this package doesn't even build anymore. Fix it if you want this PR to be merged.

PR #411, opened 2026-08-19
sidedata: bump package to 1.4.3

matthane:
Updates the sidedata addon to v1.4.3. The bundled libdovi moves from 3.3.1 to 3.4.0 and is now built with upstream's release-deploy profile, matching quietvoid's official binaries. Parser output is unchanged and the addon was tested on device.


Portisch:
Is libdovi 3.4.0 required or does it still work with current version?
Because then we need to bump this first.

matthane:
There is no compatibility issue or conflict between versions 3.3.1 and 3.4.0, so you can bump CE's version if you want, but it will still work just fine with the module using its bundled 3.4.0 regardless. 3.4.0 just has some input hardening but output is the same.

PR #415, opened 2026-09-13
aml_bt.service: Do NOT rfkill-block bluetooth for proper shutdown handling

sotux:
Do NOT rfkill-block bluetooth on stop. The W1 power-down it triggers makes the
aml_sdio/aml_w1_sdio driver retry SDIO transfers against a powered-off chip
forever (thousands of "config_pmu_reg wifi_pmu_status:0xff" and
"meson-gx-mmc fe088000.sdio: ... TIMEOUT[0x2800]" per second), which wedges
the ordered shutdown so that the box never reboots.


Portisch:
This is only about W1 driver. What about others?
Maybe the driver itself should be fixed?

sotux:
I tried to fix drivers, but it still blocked the systemd if not comment ExecStopPost line.
I have not other devices to test. 
My driver fix is here:
https://github.com/sotux/w1-aml/commit/8079fa56a7786d0af3035505dd142e8c136ed2d7
https://github.com/sotux/aml_bt/commit/6b0efbf473cbad458d4248d5f32c72230c39e70a
I am not familiar with hardware driver and these fix codes are assistant by AI.

